PR #47, opened 2020-04-08
Remove gcs range download

chrieke:
- Removes the range header download (actually not the solution for gcs inconsistencies).
- Uses chunks.
- Adds (skipped) huge download to test gcs 2min treshold (test successfull).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
-mkdocs==1.0.4
+mkdocs==1.1
 mkdocs-material
 pandoc
 pygments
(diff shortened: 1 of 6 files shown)


chrieke:
> Nice!

Thanks for the discussion and help!

pmihal-up42:
👍 good job

chrieke:
> Is this the solution for the download problems you encountered yesterday?

Not completely. We figured out with Peters help that the ranged bytes download is actually definitely not the solution so removing that. The incosistencies are gone right now and we added a test, but we need to monitor it and might have to come back. Maybe the chunked download also helps.

PR #52, opened 2020-04-15
ignore gh-pages

chrieke:
Ignores github-pages branch in Circleci (docs branch that only contains the hmtl pages).

https://circleci.com/docs/2.0/configuration-reference/#full-example

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,8 +24,8 @@ workflows:
   version: 2
   build-and-deploy:
     jobs:
-      - build_and_test
-        filters:
-          branches:
-            ignore:
-              - gh-pages
+      - build_and_test:
+          filters:
+            branches:
+              ignore:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -21,4 +24,8 @@ workflows:
   version: 2
   build-and-deploy:
     jobs:
-      - build_and_test
+      - build_and_test:
+          filters:
```
markusuwe, line 28: Is this necessary when using the ignore: above? Did you try without?

chrieke: k, ignoring the gh-pages branch does not work, as it would require a ci file in the branch (but handled by mkdocs automatically so no option to add that). Now just skipping for the docs ci by adding "[ci skip]" to the commit message (make gh-pages). Thanks for looking into this!

PR #57, opened 2020-04-26
Simplify accepted sensors in catalog search

chrieke:
- Simplify accepted sensors&blocks logic for catalog search
- fix sensor name sentinel5p

### DIFF
--- a/up42/catalog.py
+++ b/up42/catalog.py
@@ -18,28 +18,21 @@ logger = get_logger(__name__)
 # TODO: Midterm add catalog results class? Scenes() etc. that also as feedback to workflow input.
 # Scenes() would be dataframe with quicklook preview images in it.
 
-
-blocks_default = [
-    "oneatlas-pleiades-fullscene",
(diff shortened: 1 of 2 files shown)


nekhtiari:
LGTM

up42/catalog.py
```diff
@@ -18,28 +18,21 @@
...
+        "sobloo-sentinel1-l1c-grd-aoiclipped",
+        "sobloo-sentinel1-l1c-slc-full",
+    ],
+    "sentinel2": [
+        "sobloo-sentinel2-lic-msi-full",
+        "sobloo-sentinel2-lic-msi-aoiclipped",
+    ],
+    "sentinel3": ["sobloo-sentinel3-full"],
+    "sentinel5p": ["sobloo-sentinel5-preview-full",],
+}
```
rodrigoalmeida94, line 35: Nice! Would be great if we could fetch this info from some endpoint on the backend (maybe specs)?

chrieke: Good idea, added to feature roadmap.

PR #60, opened 2020-04-26
Add test_job

chrieke:
Splits `create_and_run_job` into `run_job` and `test_job` (alternative .query() or test_query() but less clear). Makes it easier to discover and safer to use (feedback!).  Breaking change due to renaming, but simplification justifies it imho and still possible right now, started the changelog.
- Adjusted in cli as well
- Adds test
- Adjusts docs


nekhtiari:
Nice

up42/cli.py
```diff
@@ -436,19 +436,30 @@ def add_workflow_tasks(workflow, input_tasks_json):
...
-    "--test-query",
-    help="Only search for available imagery based on your data parameters",
-    is_flag=True,
-)
 @click.option("--track", help="Track status of job in shell.", is_flag=True)
 @click.pass_obj
-def create_and_run_job(workflow, input_parameters_json, test_query, track):
+def test_job(workflow, input_parameters_json, track):
+    """
+    Creates and runs a new job.
```
rodrigoalmeida94, line 443: Maybe add something here 
> Create a run a new test job  (Test Query). With this test query you will not be charged with any data or processing credits, but have a preview of the job result.

up42/_version.txt
```diff
@@ -1 +1 @@
-0.8.1b7
\ No newline at end of file
+0.8.2b0
```
markusuwe, line 1: Unusual versioning. Why do we use the "bX" at the end?

up42/cli.py
chrieke: Thanks, added also to other docstrings & docs examples!

up42/_version.txt
chrieke: Will switch to 0.8.2

PR #66, opened 2020-04-30
Avoid provider arg catalog quicklook

chrieke:
- Instead of the image `provider` the user selects the `sensor` as argument for download_quicklooks. Much easier as the sensor needs to be specified for catalog search anyway and the user might not know about the provider. 
- Also removes the backend string as not required.

### DIFF
--- a/up42/__init__.py
+++ b/up42/__init__.py
@@ -41,11 +41,11 @@ def initialize_project() -> "Project":
     return Project(auth=_auth, project_id=_auth.project_id)
 
 
-def initialize_catalog(backend: str = "ONE_ATLAS") -> "Catalog":
+def initialize_catalog() -> "Catalog":
     """Directly returns a Catalog object."""
     if _auth is None:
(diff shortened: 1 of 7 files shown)


up42/__init__.py
```diff
@@ -41,11 +41,11 @@ def initialize_project() -> "Project":
     return Project(auth=_auth, project_id=_auth.project_id)
 
 
-def initialize_catalog(backend: str = "ONE_ATLAS") -> "Catalog":
+def initialize_catalog() -> "Catalog":
```
rodrigoalmeida94, line 44: So the ONEATLAS thing is not required anymore by catalog search?

chrieke: As far as I can see it is not required, will confirm with backend on Monday. Examples with and without it in postman, but does not seem to make a difference, without & without it same: possible combinations of sensors, number of images, effect of limit parameter.

PR #67, opened 2020-05-01
add settings for logging

chrieke:
Disables/Enables logging output (can be a lot in the notebooks)

### DIFF
--- a/up42/__init__.py
+++ b/up42/__init__.py
@@ -73,6 +73,20 @@ def initialize_jobtask(jobtask_id, job_id) -> "JobTask":
     )
 
 
+def control_logging(status=False):
+    if not status:
+        logger.info(
+            "Disabling logging! Use up42.set_logging(status=True) to reactivate."
+        )
+    else:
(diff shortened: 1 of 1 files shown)


chrieke:
> Nice! Maybe also something to add to the CLI? Like the verbose/silent mode

@rodrigoalmeida94 Good idea, with click everything needs to be bound at least to the auth objects that persists correct? So would need to move this to Tools and then use it like get block?

rodrigoalmeida94:
> > Nice! Maybe also something to add to the CLI? Like the verbose/silent mode
> 
> @rodrigoalmeida94 Good idea, with click everything needs to be bound at least to the auth objects that persists correct? So would need to move this to Tools and then use it like get block?

Well maybe the logger object can be a class instance variable? That way you can access it from wherever, might involve some refactoring to do this in the cli

chrieke:
@rodrigoalmeida94 thanks! Okay added this to the roadmap for next iteration, will merge for now  without cli option.

PR #72, opened 2020-05-06
Get jobs for projects

chrieke:
- Adds Project().get_jobs() as getjob is actually unrelated to workflow and should be accessible at this level.
- Adds a filter to Workflow.get_jobs() to only return the jobs that are associated with the workflow object.

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -1,7 +1,7 @@
 import requests_mock
 import pytest
 
-from .context import Project, Workflow
+from .context import Project, Workflow, Job
 
 # pylint: disable=unused-import
 from .fixtures import (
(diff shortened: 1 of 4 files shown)


tests/test_project.py
```diff
@@ -113,6 +113,38 @@ def test_get_workflows_live(project_live):
...
+        )
+        m.get(url=url_job_info, json={"data": {"xyz": 789}, "error": {}})
+
+        jobs = project_mock.get_jobs()
+        assert isinstance(jobs, list)
+        assert isinstance(jobs[0], Job)
+        assert jobs[0].job_id == job_id
+
+
+@pytest.mark.skip
```
markusuwe, line 139: Left over or on purpose?

PR #74, opened 2020-05-07
Variable job name

chrieke:
Adds job name option. By default takes workflow_name (as in the console) and adds _py

### DIFF
--- a/up42/workflow.py
+++ b/up42/workflow.py
@@ -351,6 +351,7 @@ class Workflow(Tools):
         input_parameters: Union[Dict, str, Path] = None,
         test_job=False,
         track_status: bool = False,
+        name: str = None,
     ) -> "Job":
         """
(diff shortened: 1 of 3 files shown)


up42/workflow.py
```diff
@@ -383,7 +385,9 @@ def _helper_run_job(
 
         logger.info("Selected input_parameters: %s.", input_parameters)
 
-        name = "_py"  # Enables recognition of python API usage.
+        if name is None:
+            name = self.info["name"]
+        name = f"{name}_py"  # Temporary recognition of python API usage.
```
rodrigoalmeida94, line 390: Is this still needed given we are already passing the custom headers in every call?

chrieke: Unfortunately yes, backend + bi solution is not yet completely figured out. We can just remove later without breaking anything.

PR #82, opened 2020-05-11
Sdk-notebook/up-4389

marooned20:
Following notebook submission contains example of using up42 sdk with additional computation. It is directed toward using Modified NDWI to map flood impacted areas using building footprints from OSM!


marooned20:
> Awesome @marooned20! One suggestion I have is you can maybe try to use the [NDWI block from our partner](https://marketplace.up42.com/block/6559a2d2-f023-461a-a14d-f094618c9ba0) instead of computing it locally - I think that can be a better showcase of capabilities.

Ahh, I guess I'm blind really! I couldn't find it earlier. Let me make changes in that case!

nekhtiari:
Nice explorative job!

marooned20:
@markusuwe 
Yes, I am not mistaken this issue comes from underlying libspatialindex (subsequently Rtree) not being installed. A `brew install spatialindex` and `sudo apt install libspatialindex-dev` (considering *nix based dockerizing) should do the trick!
> Normalisation function can be replaced by np.linalg.norm(v, ord=1) 
Will do!

marooned20:
Couple of points:
- normalisation strategy works best with current implementation. I tried with `sklearn.preprocessing` startegies but none seem to produce the expected result.
- Regarding using partner block, this one calculates Modified NDWI which is rather sensitive to the water compared to regular NDWI. This index uses SWIR and green channel combination instead of NIR one. which in return effectively detects water and reduces the false signal.

marooned20:
@chrieke I pushed notebook with cosmetic and minor syntactic changes as suggested.
- <del> shapely install should not be required (dependency of geopandas)
- <del> Maybe do something like “can be installed directly in the notebook via: !pip install osmnx”
- <del> Maybe briefly mention in the intro that this notebook shows that the up42 sdk can be seamlessly integrated in a geospatial workflow in Python or geospatial python stack etc (in better words 
- <del> Can you check the imports os, pandas, rasterio features maybe more not required
- <del> np.seterr(divide=‘ignore’, invalid=‘ignore’)   maybe add directly under np import
> - <del> Would maybe not use data_dir, doesnt really add anything (default “None” argument creates a project_projectid folder structure </del> 
> Having specified custom folder is just for showing the otherwise. But put an additional note that default is also possible
- <del> in this case maybe put the normalize function directly above cell where it is used, in the beginning the user should not be distracted from sdk usage
> - <del> The “getting actual paths” from job.results loop is interesting. Looks a bit awkward but yeah kinda neccessary, maybe we should instead add the json path to job.json and images to job.images etc in the sdk. Maybe in this case as you use the path only once and it is one image anyway, just do   “rio.open(job.results[0])” or use endswith here, same for json. Certainly not good Python but would look way less complex without the awkward loop. </del> 
> As discussed on slack earlier, this is required due to the fact that it does not necessarily return the files in deterministic order. However, added a more pythonic way to extract the paths
- <del> Remove or put a few of the “shape” “head” etc. prints in the same cell as the variable, saves a bit of space/focuses on meaningful action. In general try to make it a bit compact maybe
maybe 1 less quick plots, you show the vectors anyway later
> - <del> love the color on the plot, we should put a screenshot gallery one for each example notebook in the docs examples main page </del> 
> I looked into possible solutions to export folium map as a png. At the moment folium does not support it out of the box. Although there are some hacky ways suggested [here](https://github.com/python-visualization/folium/issues/35). But it just adds overhead for the showcase example purpose. We could take a less painful screenshot path instead! 

> - <del> effected.to_crs(crs=‘EPSG:4326’, inplace=True)    maybe assign it to effected, we try to avoid inplace (think there are also plans to deprecate it in geopandas) </del> 
> This is necessary because `effected` dataframe is created by query filter and it's not possible to declare epsg directly

PR #85, opened 2020-05-14
Avoid potential .map_results error

chrieke:
- Avoid potential error where the data.json was selected for plotting the raster file.
- Pulls out read_raster_in_4326 into utils


dc-up42:
Nice catch.

> Avoid potential error by non-deterministic results file order

I may be mistaken, but it doesn't look like you sort the file paths in any way to avoid the non-deterministic result.

chrieke:
> I may be mistaken, but it doesn't look like you sort the file paths in any way to avoid the non-deterministic result.

Thank you, yep, ticket still open.

Makefile
```diff
@@ -1,7 +1,7 @@
 SRC := .
 
 env:
-	mkvirtualenv --python=$(which python3.7) up42-py
+	mkvirtualenv --python=$(which python3) up42-py
```
markusuwe, line 4: Are we sure the python sdk works with all python 3 versions?

up42/utils.py
```diff
@@ -421,3 +424,53 @@ def fc_to_query_geometry(
...
+            dst_profile.update(
+                {
+                    "crs": dst_crs,
+                    "transform": transform,
+                    "width": width,
+                    "height": height,
+                }
+            )
+
+            with MemoryFile() as memfile:
```
marooned20, line 460: Nice usage of memoryfile! 💯

Makefile
chrieke: >3.6, but seems there is no way to use `which` like that. Any idea @marooned20? Otherwise will just use `which python3.8` to indicate the newest version.

marooned20: if the prerequisite to run this is `>= python3.6` then it's better to use a pinned version of python (either of 3.6, 3.7 or 3.8) instead of just python3.

PR #101, opened 2020-07-08
Add parameters mapping

rodrigoalmeida94:
This is the first out of three (1/3) PRs intended to provide the parallel job run feature in the python SDK. We will work on the feature branch `add-parallel-jobs` and only when after all 3 PRs are merged we will make a PR to master for deployment.

- [x] Adds JobCollection class (for now not used, will be returned when parallel job run is called)
- [x] Add mapping to parameters function

Handles three cases:

1. Mapping n x Scene IDs with n x Geometries
2. Mapping n x Time Intervals with n x Geometries
3. Mapping n x Scene IDs

Requires merge of https://github.com/up42/up42-py/pull/102 for passing tests.

### DIFF
--- a/up42/workflow.py
+++ b/up42/workflow.py
@@ -2,7 +2,7 @@ import json
 import logging
 from collections import Counter
 from pathlib import Path
-from typing import Dict, List, Union, Optional
+from typing import Dict, List, Union, Optional, Tuple
 
 from geopandas import GeoDataFrame
 from shapely.geometry import Point, Polygon
@@ -346,18 +346,34 @@ class Workflow(Tools):
         return input_parameters
 
     def mapping_to_parameters(
-        self, geometries: List[Feature], time_series: List[str]
+        self,
+        geometries: List[
+            Union[
+                Dict,
+                Feature,
+                FeatureCollection,
+                geojson_Polygon,
+                List,
+                GeoDataFrame,
+                Polygon,
+                Point,
+            ]
+        ],
+        time_series: List[Tuple[str, str]],
+        limit_per_job: int = 1,
+        geometry_operation: str = "intersects",
     ) -> List[dict]:
         result_params = []
         for geo in geometries:
             # TODO: Fix time_series handling
-            for time in time_series:
+            for start_date, end_date in time_series:
                 result_params.append(
                     self.construct_parameters(
                         geometry=geo,
-                        geometry_operation="intersects",
-                        start_date=time,
-                        end_date=time,
+                        geometry_operation=geometry_operation,
+                        start_date=start_date,
+                        end_date=end_date,
+                        limit=limit_per_job,
                     )
                 )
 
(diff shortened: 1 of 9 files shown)


nekhtiari:
Nice! I would also like to discuss about this as well tmr in our duos.

up42/workflow.py
```diff
@@ -342,6 +342,43 @@ def construct_parameters(
             input_parameters[data_block_name][geometry_operation] = aoi_feature
         return input_parameters
 
+    def mapping_to_parameters(
+        self,
+        geometries: List[Union[Dict, Feature, geojson_Polygon, Polygon, Point,]],
+        interval_dates: List[Tuple[str, str]],
```
chrieke, line 348: Would it be better to start with `image_ids` instead of `interval_dates` here to make it easier to grasp? Think this was suggested in the discussion. Because date handling in the sdk is still lacking (missing `time_series` in `construct_parameters`, accepted date formats should be extended) this would also keep our options open to  then just add `time_series` here.

up42/workflow.py
```diff
@@ -342,6 +342,43 @@ def construct_parameters(
             input_parameters[data_block_name][geometry_operation] = aoi_feature
         return input_parameters
 
+    def mapping_to_parameters(
```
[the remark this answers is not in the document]
rodrigoalmeida94: > 

Maybe I'm missing something, why would we make this method private? I expect users to make use of it directly.

up42/workflow.py
rodrigoalmeida94: The scope for this was pretty clear in the point were we want to support multiple geoms and multiple timestamps  i.e. 
```
1. Same AOI, multiple timesteps
- limit N → N datasets → N jobs
2. Different AOIs, same/different timesteps 
- N AOIs → N datasets → N Jobs
```
Are you hinting that the user would always then do the catalog search and then run this mapping?

nekhtiari: I also thought the same. Since this will affect the parallelization as well, I guess we need to be clear what exactly we want.

up42/workflow.py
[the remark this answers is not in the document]
chrieke: Makes sense, thought would be used by Nikoo later.

up42/workflow.py
chrieke: A okay, think we misscommunicated. Thought about timesteps as multiple scenes, did not consider the time periods we use via `time_series` parameter. @markusuwe also suggested the catalog output scene_ids as the input I think, maybe I missed the conclusion. How would parallelizing multiple scene_ids be handled, would that need to be added here?

Would `construct_parameters_parallel` a more descriptive name?

rodrigoalmeida94: Changed name of method and added three cases as discussed.

nekhtiari:
LGTM

PR #102, opened 2020-07-09
Allow passing of only scene_ids in construct_parameters

rodrigoalmeida94:
We found in https://github.com/up42/up42-py/pull/101 that the construct parameters is currently not supporting the creation of job parameters that include only scene ids (without a geometry). This change ensures that the parameters are generated even when no geometry is passed.


nekhtiari:
LGTM

PR #104, opened 2020-07-14
buffer 0 in any_vector_to_fc to avoid self intersections

chrieke:
buffer 0 in any_vector_to_fc to avoid self intersections.


nekhtiari:
Nice and clean!

chrieke:
> LGTM
> That's extremely simple.
> @chrieke and @marooned20 Can't we do the exact same in the blocks (e.g. OneAtlas) to resolve the problems in https://up42.atlassian.net/browse/UP-4647 ?

Think so, let's discuss offline / when Prayag is back, set a reminder.

PR #114, opened 2020-07-28
Merge jobcollection results

rodrigoalmeida94:
- Adds flag (defaults to True) to download_results in order to merge all data.json from a JobCollection into a single data.json with updated `up42.data_path`
- Keep folder structure when unpacking results from tar file
- Adds tests for unpacked result download and DIMAP file download
- Ignore not-callable temporarily

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -135,7 +135,7 @@ def jobs_mock(auth_mock):
 
 
 @pytest.fixture()
-def jobcollection_mock(auth_mock, job_mock):
+def jobcollection_single_mock(auth_mock, job_mock):
     return JobCollection(
         auth=auth_mock, project_id=auth_mock.project_id, jobs=[job_mock]
     )
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -114,9 +114,37 @@ def job_mock(auth_mock):
 
 
 @pytest.fixture()
+def jobs_mock(auth_mock):
+    with requests_mock.Mocker() as m:
+        job_id = "jobid_123"
+        url_job_info = (
+            f"{auth_mock._endpoint()}/projects/{auth_mock.project_id}/jobs/{job_id}"
+        )
+        m.get(url=url_job_info, json={"data": {"xyz": 789}, "error": {}})
+
+        job1 = Job(auth=auth_mock, project_id=auth_mock.project_id, job_id=job_id)
+
+        job_id = "jobid_456"
+        url_job_info = (
+            f"{auth_mock._endpoint()}/projects/{auth_mock.project_id}/jobs/{job_id}"
+        )
+        m.get(url=url_job_info, json={"data": {"xyz": 789}, "error": {}})
+
+        job2 = Job(auth=auth_mock, project_id=auth_mock.project_id, job_id=job_id)
+    return [job1, job2]
+
+
+@pytest.fixture()
 def jobcollection_mock(auth_mock, job_mock):
     return JobCollection(
         auth=auth_mock, project_id=auth_mock.project_id, jobs=[job_mock]
+    )
+
+
+@pytest.fixture()
+def jobcollection_multiple_mock(auth_mock, jobs_mock):
+    return JobCollection(
+        auth=auth_mock, project_id=auth_mock.project_id, jobs=jobs_mock
     )
 
 
--- a/tests/test_jobcollection.py
+++ b/tests/test_jobcollection.py
@@ -2,6 +2,7 @@ from pathlib import Path
 import tempfile
 
 import pytest
+import requests_mock
 
 # pylint: disable=unused-import,wrong-import-order
 from .fixtures import (
@@ -16,6 +17,30 @@ from .fixtures import (
 
 def test_jobcollection(jobcollection_mock):
     assert len(jobcollection_mock.jobs) == 1
+
+
+def test_jobcollection_download_result(jobcollection_mock):
+    with requests_mock.Mocker() as m:
+        download_url = "http://up42.api.com/abcdef"
+        url_download_result = (
+            f"{jobcollection_mock.auth._endpoint()}/projects/"
+            f"{jobcollection_mock.project_id}/jobs/{jobcollection_mock.jobs_id[0]}/downloads/results/"
+        )
+        m.get(url_download_result, json={"data": {"url": download_url}, "error": {}})
+
+        out_tgz = Path(__file__).resolve().parent / "mock_data/result_tif.tgz"
+        out_tgz_file = open(out_tgz, "rb")
+        m.get(
+            url=download_url,
+            content=out_tgz_file.read(),
+            headers={"x-goog-stored-content-length": "163"},
+        )
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            out_dict = jobcollection_mock.download_results(tempdir)
+            for job_id in out_dict:
+                assert Path(out_dict[job_id][0]).exists()
+                assert len(out_dict[job_id]) == 2
 
 
 @pytest.mark.live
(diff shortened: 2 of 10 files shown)


up42/jobcollection.py
```diff
@@ -72,5 +72,29 @@ def download_results(
...
+                    all_files = out_filepaths[job_id]
+                    data_json = [d for d in all_files if d.endswith("data.json")][0]
+                    with open(data_json) as src:
+                        data_json_fc = geojson.load(src)
+                        for feat in data_json_fc.features:
+                            feat.properties["job_id"] = job_id
+                            try:
+                                feat.properties[
+                                    "up42.data_path"
+                                ] = f"job_{job_id}/{feat.properties['up42.data_path']}"
```
chrieke, line 89: nice

tests/fixtures.py
```diff
@@ -113,13 +113,41 @@ def job_mock(auth_mock):
...
+
 @pytest.fixture()
 def jobcollection_mock(auth_mock, job_mock):
     return JobCollection(
         auth=auth_mock, project_id=auth_mock.project_id, jobs=[job_mock]
     )
 
 
+@pytest.fixture()
+def jobcollection_multiple_mock(auth_mock, jobs_mock):
```
chrieke, line 145: Can you rename the above `jobcollection_mock` to `jobcollection_single_mock` please. Pretty sure I would grab the wrong one thinking this naturally has multiple.

tests/test_jobcollection.py
```diff
@@ -18,10 +22,86 @@ def test_jobcollection(jobcollection_mock):
     assert len(jobcollection_mock.jobs) == 1
 
 
+def test_jobcollection_multiple(jobcollection_multiple_mock):
+    assert len(jobcollection_multiple_mock.jobs) == 2
+
+
+def test_jobcollection_download_result(jobcollection_mock):
```
chrieke, line 29: Test names always say "_result" instead of "_results" like the function. Not strict but would be good to fix in case someone searches for the test.

tests/test_jobcollection.py
```diff
@@ -18,10 +22,86 @@ def test_jobcollection(jobcollection_mock):
...
+    with requests_mock.Mocker() as m:
+        download_url = "http://up42.api.com/abcdef"
+        url_download_result = (
+            f"{jobcollection_mock.auth._endpoint()}/projects/"
+            f"{jobcollection_mock.project_id}/jobs/{jobcollection_mock.jobs_id[0]}/downloads/results/"
+        )
+        m.get(url_download_result, json={"data": {"url": download_url}, "error": {}})
+
+        out_tgz = Path(__file__).resolve().parent / "mock_data/result_tif.tgz"
+        out_tgz_file = open(out_tgz, "rb")
```
marooned20, line 39: Suggestion: Context manager here would prevent file being left open in memory after test has run

tests/test_jobcollection.py
```diff
@@ -18,10 +22,86 @@ def test_jobcollection(jobcollection_mock):
...
+        m.get(
+            url=download_url,
+            content=out_tgz_file.read(),
+            headers={"x-goog-stored-content-length": "163"},
+        )
+
+        with tempfile.TemporaryDirectory() as tempdir:
```
marooned20, line 75:
```suggestion
        with tempfile.TemporaryDirectory() as tmpdir:
```
tempfile package contains `tempdir` method, just to prevent any conflict with in-built function naming!

tests/test_jobcollection.py
```diff
@@ -18,10 +22,86 @@ def test_jobcollection(jobcollection_mock):
...
+            out_dict = jobcollection_mock.download_results(tempdir, merge=False)
+            for job_id in out_dict:
+                assert Path(out_dict[job_id][0]).exists()
+                assert len(out_dict[job_id]) == 2
+
+
+def test_jobcollection_download_result_merged(jobcollection_multiple_mock):
+    with requests_mock.Mocker() as m:
+        download_url = "http://up42.api.com/abcdef"
+        url_download_result_1 = (
```
marooned20, line 56: Cosmetic suggestion: Here a for loop could reduce the code by half!

tests/test_project.py
```diff
@@ -174,4 +174,5 @@ def test_get_project_settings_live(project_live):
     project_settings = project_live.get_project_settings()
     assert isinstance(project_settings, list)
     assert len(project_settings) == 3
-    assert project_settings[0]["name"] == "MAX_CONCURRENT_JOBS"
+    project_settings_dict = {item["name"]: item for item in project_settings}
+    assert "MAX_CONCURRENT_JOBS" in project_settings_dict.keys()
```
marooned20, line 178: Nice!

up42/jobcollection.py
```diff
@@ -72,5 +72,29 @@ def download_results(
             )
             out_filepaths[job.job_id] = out_filepaths_job
 
+        if merge:
+            merged_data_json = output_directory / "data.json"
+            with open(merged_data_json, "w") as dst:
+                out_features = []
+                for job_id in out_filepaths:
+                    all_files = out_filepaths[job_id]
+                    data_json = [d for d in all_files if d.endswith("data.json")][0]
```
marooned20, line 81: is there a possibility that the folder may contain a file, e.g. `metadata.json` which gets captured by this?

tests/fixtures.py
rodrigoalmeida94: Good point, changed.

tests/test_jobcollection.py
rodrigoalmeida94: Fixed

tests/test_jobcollection.py
rodrigoalmeida94: Good catch! Changed

tests/test_jobcollection.py
rodrigoalmeida94: Right, didn't think of this, changed!

tests/test_jobcollection.py
rodrigoalmeida94: Good suggestion, applied

up42/jobcollection.py
rodrigoalmeida94: It's a good question, I would guess it would get captured, and we don't want that. Changed to use the Path name property to make it more precise.

nekhtiari:
Nice, LGTM

up42/utils.py
```diff
@@ -71,18 +72,17 @@ def download_results_from_gcs(
...
-        for file in files:
-            f = tar.extractfile(file)
-            content = f.read()  # type: ignore
-            out_fp = output_directory / f"{Path(file.name).name}"
-            with open(out_fp, "wb") as dst:
-                dst.write(content)
-            out_filepaths.append(str(out_fp))
+        tar.extractall(path=output_directory)
+        output_folder_path = output_directory / "output"
+        for src_path in output_directory.glob("output/**/*"):
```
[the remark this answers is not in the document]
rodrigoalmeida94: So the try except block is there because for some reason, globbing the output folder returns the output directory in this list - this just make sure all the files are moved (I checked this again and removed it, seems the error I got before is gone). The solution you propose would not work since you actually have to move files and folders around, not just rename them/change the paths. Also the solution in SO we would come into the same problem of not keeping the folder structure.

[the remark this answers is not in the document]
chrieke: 👍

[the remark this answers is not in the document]
chrieke: Not sure if deliberatly globbing `output_directory.glob("output/**/*")`, but maybe the previous issue with the directory would be even saver with globbing in the folder directly instead of using output in the pattern: `output_folder_path.rglob("*")`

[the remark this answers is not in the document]
rodrigoalmeida94: Good point, will change.

PR #115, opened 2020-07-30
JobCollection fail handling

rodrigoalmeida94:
When running jobs in parallel, JobCollection will always be returned, even if there are failed jobs within the collection.

- [x] JobCollection is subscriptable and iteratable 
- [x] Adds property is_succeeded to Job
- [x] Adds get_jobs_status and get_jobs_info to JobCollection
- [x] download_results of JobCollection only downloads succeeded jobs, raise ValueError if all jobs are failed.

### DIFF
--- a/tests/test_jobcollection.py
+++ b/tests/test_jobcollection.py
@@ -5,6 +5,8 @@ import geojson
 import pytest
 import requests_mock
 
+from .context import Job
+
 # pylint: disable=unused-import,wrong-import-order
 from .fixtures import (
     auth_mock,
@@ -95,6 +97,11 @@ def test_jobcollection_download_results_merged(jobcollection_multiple_mock):
                     tmpdir
                     / Path(merged_data_json.features[0].properties["up42.data_path"])
                 ).exists()
+
+
+def test_jobcollection_subscripted(jobcollection_single_mock):
+    assert isinstance(jobcollection_single_mock[0], Job)
+    assert jobcollection_single_mock[0].job_id == "jobid_123"
 
 
 @pytest.mark.live
--- a/tests/test_jobcollection.py
+++ b/tests/test_jobcollection.py
@@ -30,15 +30,13 @@ def test_jobcollection_multiple(jobcollection_multiple_mock):
 
 def test_job_iterator(jobcollection_multiple_mock):
     worker = lambda job: 1
-    res = jobcollection_multiple_mock._jobs_iterator(worker, only_succeeded=False)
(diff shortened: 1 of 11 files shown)


tests/test_jobcollection.py
```diff
@@ -26,15 +29,82 @@ def test_jobcollection_multiple(jobcollection_multiple_mock):
     assert len(jobcollection_multiple_mock.jobs) == 2
 
 
+def test_job_iterator(jobcollection_multiple_mock, jobcollection_empty_mock):
+    worker = lambda job: 1
```
marooned20, line 33: clean and clever!

up42/jobcollection.py
```diff
@@ -33,10 +33,69 @@ def __repr__(self):
...
 
-    # TODO: Make class subscriptable.
+    def __getitem__(self, index: int) -> Job:
+        return self.jobs[index]
+
+    def __iter__(self):
+        for job in self.jobs:
+            yield job
+
+    def apply(
```
chrieke, line 43: Probably private method best, also to avoid in docs. Or do you see any usage by the user outside of the provided functions?

up42/jobcollection.py
```diff
@@ -33,10 +33,69 @@ def __repr__(self):
...
+                if job.is_succeeded:
+                    out_dict[job.job_id] = worker(job, **kwargs)
+            else:
+                out_dict[job.job_id] = worker(job, **kwargs)
+        if not out_dict:
+            if self.jobs:
+                raise ValueError(
+                    "All jobs have failed! Cannot apply over an empty succeeded job list."
+                )
+            if not self.jobs:
```
chrieke, line 74: Nice functionality. A bit simpler like this maybe, at least the raises top&bottom.
```
if not self.jobs:
    raise ValueError("This is an empty JobCollection. Cannot apply over an empty job list.")

out_dict = {}
for job in self.jobs:
     if only_succeeded:
            if not job.is_succeeded:
                 continue
     out_dict[job.job_id] = worker(job, **kwargs)

if not out_dict:
   raise ValueError("All jobs have failed! Cannot apply over an empty succeeded job list.")
```

up42/jobcollection.py
rodrigoalmeida94: I think makes sense to make it public, so users could built custom loops if they want, like getting all job params for instance. We could implement more but if it's public users can also do that.

up42/jobcollection.py
rodrigoalmeida94: Very good suggestion, looks much better, applied!

PR #116, opened 2020-07-31
Flood mapping osm block

marooned20:
- [x] Configure project to use osm block


chrieke:
LGTM, multiple osmnx library references & imports should be removed though.

marooned20:
> LGTM, multiple osmnx library references & imports should be removed though.
Good catch, will update it!
 
> Please merge but don't update the docs right now, since the parallelization stuff would be included as well.
Sure thing!

PR #118, opened 2020-08-04
Preconfigured workflows

chrieke:
- Adds `.get_templates()` and `.add_template` as equivalents of `get_blocks` and `add_workflow_tasks`. 
- Adds json with workflow templates (draft)
- Adds docs page (draft, could look better in final version)

Not sure about this solution, multiple ways to go about this. Main question is how it would be used by the user in order to surface the template easily, create a workflow out of one, and make it easy to use the template specific parameters (these could deviate from the default parameters). Initially thought about a function that just runs the template, but thinking about how it would work on the console (The templates are surfaced in the marketplace, you select one, the workflow builder & parameters are prefilled), this came closes to it while retaining the flexibility.

- Could also just be integrated with `add_workflow_tasks` instead of another function.
- Could also be more complicated, e.g. adding `self.parameters` to Workflow, and setting it with `add_workflow_tasks` or `.add_template`. This way if the user uses `construct_parameters` we could pick up the already set parameters. But adds maybe unneccessary comlexity / less explicit.

image


up42/workflow.py
```diff
@@ -229,7 +235,45 @@ def add_workflow_tasks(self, input_tasks: Union[List[str], List[Dict]]) -> None:
             f"{self.workflow_id}/tasks/"
         )
         self.auth._request(request_type="POST", url=url, data=input_tasks)
-        logger.info("Added tasks to workflow: %r", input_tasks)
+        logger.info(f"Added tasks to workflow: {input_tasks}")
+
+        if return_parameters:
+            default_parameters = self._get_default_parameters()
+            logger.info(f"Returning default workflow parameters: {default_parameters}")
+            return default_parameters
```
rodrigoalmeida94, line 243: Adding an optional return value for a function feels a bit weird - is it completely necessary?

up42/workflow.py
```diff
@@ -229,7 +235,45 @@ def add_workflow_tasks(self, input_tasks: Union[List[str], List[Dict]]) -> None:
...
+        logger.info(
+            f"Add template {template_name} to workflow containing tasks: {input_tasks}"
+        )
+
+        if return_parameters:
+            template_parameters = self.get_templates(basic=False)[template_name]
+            logger.info(
+                f"Returning template workflow parameters of template {template_name}: {template_parameters}"
+            )
+            return template_parameters
```
rodrigoalmeida94, line 276: Same here. These are methods of the workflow class, if they are adding_template to workflow, this should be the only thing they do as I see it.

up42/workflow.py
chrieke: Agree, this is not the best way to do it. Was trying to provide a solution to provide the template parameters to the user directly, but the simplest solution for now is probably to provide that manually with `.get_templates()` etc. Will  remove this when revisiting the templates.

chrieke:
Closing. Can be picked up again and reworked when properly integrating the templates on the platform.

PR #119, opened 2020-08-05
Add limitation for max concurrent jobs

rodrigoalmeida94:
Currently the maximum number of concurrent jobs is a project setting in UP42. This PR makes necessary changes that check the maximum number of concurrent jobs in settings and raises a ValueError when users select a higher number in the run_parallel_job method. This avoids unhandled API errors when the number of concurrent jobs is exceeded.

In addition:

- [x] Removed relative imports from modules to avoid ImportError
- [x] Added max_concurrent_jobs property to Project
- [x] Small fixes in docs and docstrings

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -9,6 +9,7 @@ from .fixtures import (
     auth_live,
     project_mock,
     project_live,
+    project_max_concurrent_jobs,
 )
 
 
@@ -169,6 +170,12 @@ def test_get_project_settings(project_mock):
     assert project_settings[0]["name"] == "MAX_CONCURRENT_JOBS"
 
 
+def test_max_concurrent_jobs(project_mock, project_max_concurrent_jobs):
+    with project_max_concurrent_jobs(5):
+        max_concurrent_jobs = project_mock.max_concurrent_jobs
+    assert max_concurrent_jobs == 5
+
+
 @pytest.mark.live
 def test_get_project_settings_live(project_live):
     project_settings = project_live.get_project_settings()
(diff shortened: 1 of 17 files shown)


tests/test_project.py
```diff
@@ -169,10 +170,22 @@ def test_get_project_settings(project_mock):
...
     assert isinstance(project_settings, list)
     assert len(project_settings) == 3
     project_settings_dict = {item["name"]: item for item in project_settings}
     assert "MAX_CONCURRENT_JOBS" in project_settings_dict.keys()
+
+
+@pytest.mark.live
+def test_max_concurrent_jobs_live(project_live):
+    max_concurrent_jobs = project_live.max_concurrent_jobs
+    assert max_concurrent_jobs == 100
```
chrieke, line 191: Think typo, should be 10 right?

rodrigoalmeida94: Was also surprised, but if you look in the settings of the test account, it is set to 100.

chrieke: Okay interesting, good input for the discussion to increase it in general.

up42/workflow.py
```diff
@@ -694,3 +721,12 @@ def delete(self) -> None:
...
         del self
+
+    @property
+    def max_concurrent_jobs(self) -> int:
+        """Gets the maximum number of concurrent jobs allowed by the project settings."""
+        url = f"{self.auth._endpoint()}/projects/{self.project_id}/settings"
+        response_json = self.auth._request(request_type="GET", url=url)
+        project_settings = response_json["data"]
+        project_settings_dict = {d["name"]: int(d["value"]) for d in project_settings}
+        return project_settings_dict["MAX_CONCURRENT_JOBS"]
```
rodrigoalmeida94, line 732: Added this duplication to avoid circular imports from Project. We will need to fix later potentially with a ProjectSettings object.

PR #120, opened 2020-08-05
Various small improvements

chrieke:
- Makes handling of catalog search query parameter/sorting clearer with sentinel-1 (no cloudcover)
- Uses fstrings for all logging
- Removes "Optional" types for parameters (supposed to be used for optional returns)
- various small improvements see commit messages
- Removes most todos (some added to Jira)

### DIFF
--- a/up42/catalog.py
+++ b/up42/catalog.py
@@ -126,8 +126,8 @@ class Catalog(Tools):
             sort_order = "desc"
 
         query_filters = {"dataBlock": {"in": block_filters}}
-        if not sensors == ["sentinel1"]:
-            query_filters["cloudCoverage"] = {"lte": max_cloudcover}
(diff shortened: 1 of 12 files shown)


up42/catalog.py
```diff
@@ -129,15 +119,22 @@ def construct_parameters(
             squash_multiple_features="footprint",
         )
 
-        # TODO: cc also contains nan with sentinel 1 etc. ignore?
+        if ascending:
+            sort_order = "asc"
```
marooned20, line 123: Minor Suggestion: `sort_order = "asc" if ascending else "desc"`

chrieke: Not the biggest fan of oneliners but here it seems fitting, thanks

PR #122, opened 2020-08-07
Docs improvements

chrieke:
- Ignores the init function in the code reference & adds the class description as pure markdown, looks much better.
- Adds jobcollection etc.
- Adds a bit more width to the site.

### DIFF
--- a/docs/structure.md
+++ b/docs/structure.md
@@ -2,8 +2,9 @@
 
 ## Hierachy
 
-- The Python SDK uses six object classes, representing the **hierarchical structure of UP42**:
+- The Python SDK uses seven object classes, representing the **hierarchical structure of UP42**:
     - **Project > Workflow > Job > JobTask**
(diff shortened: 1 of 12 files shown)


docs/structure.md
```diff
@@ -2,8 +2,9 @@
 
 ## Hierachy
```
rodrigoalmeida94, line 3: I understand why you have this page as an "index" but could there be a way to automate this? I feel it would be very easy to miss methods/forget to update.

chrieke: Could be! The [macros plugin](https://mkdocs-macros-plugin.readthedocs.io/en/latest/) lets you define arbitrary Python functions that you can invoke in markdown like this: `{{ mymacro(...) }}`    Already tested it, works super well. Can even produce plots etc.    Not 100% sure if it would work here, but think so. Would need to get a list of the functions and return it formatted like this.

PR #129, opened 2020-09-07
Remove buffer 0

chrieke:
Now that geometry validation error messaging is integrated on data block level, removes buffer 0 application in the SDK to keep geometry handling consistent and not run into edge cases (@marooned20 had such an example today, thanks for your help!). 
We could potentially integrate the same validation messaging right in the Python code, topic is still being discussed.

Also lots of new formatting with updated version of black.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,5 @@
 mkdocs
 mkdocs-material
-pandoc
 pygments
 pymdown-extensions
 mkdocs-autolinks-plugin
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -32,7 +32,8 @@ def auth_mock():
     with requests_mock.Mocker() as m:
         url_token = "https://project_id123:project_apikey123@api.up42.com/oauth/token"
         m.post(
-            url=url_token, json={"data": {"accessToken": "token_789"}},
+            url=url_token,
+            json={"data": {"accessToken": "token_789"}},
         )
         auth = Auth(
             project_id="project_id123",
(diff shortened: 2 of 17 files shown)


marooned20 (approved):
LGTM!

markusuwe (approved):
LGTM
So many formatting changes, it was hard to find the actual relevant line. ;-)

rodrigoalmeida94 (approved):
LGTM

nekhtiari (approved):
I had a discussion with @rodrigoalmeida94 yesterday about this, and I was thinking that it is good we have consistent behavior on the SDK and platform, but on the other hand, @rodrigoalmeida94 mentioned, we do more stuff on SDK, for instance in SDK you can give different types for your geometry (Dict, Feature, FeatureCollection, List, GeoDataFrame, Polygon, Point), which is not possible in the UI, So now I am thinking, wouldn't it make sense to solve the geometry problem in the SDK but providing a good message? or in UI when the job fails, can we point to use SDK to solve this? Or we do not want to go on this road?

nekhtiari:
> I had a discussion with @rodrigoalmeida94 yesterday about this, and I was thinking that it is good we have consistent behavior on the SDK and platform, but on the other hand, @rodrigoalmeida94 mentioned, we do more stuff on SDK, for instance in SDK you can give different types for your geometry (Dict, Feature, FeatureCollection, List, GeoDataFrame, Polygon, Point), which is not possible in the UI, So now I am thinking, wouldn't it make sense to solve the geometry problem in the SDK but providing a good message? or in UI when the job fails, can we point to use SDK to solve this? Or we do not want to go on this road?

But I guess the current solution you have now might be a good solution, probably two different behaviour cause a lot of confusions ...

chrieke:
@nb-42 Yeah, behaviour now would be the same. Behaviour after removing the .buffer(0) is now that it will fail when running the block, just like on the console. What @marooned20 and I discussed yesterday is that we can add the same geometry validation directly to the SDK Python code as well. So that the user gets the same message "invalid geometry" etc., but right away before running the block & the geometry is not altered by us. See SDK maintenance Jira epic.

PR #131, opened 2020-09-09
Fix timestrings up4835

chrieke:
To be merged only catalog search api uses `+00:00` by default, or needs to be reverted in this PR for catalog search.


chrieke:
Closing since API will accept both `Z` and `+00:00`.

PR #132, opened 2020-09-09
Simpler workflow creation - up4765

chrieke:
Adds the option to define blocks to be added to the workflow right in `project.create_workflow`. A suggestion by @jngo, this would be the new default, simpler way to define a workflow. Removes a bit of clutter from all examples and imho makes the process easier. Also prepares for the simple usage of recipes/shortcuts/templates.

Think it's really neat but indeed a draft PR, open for discussion!

```
workflow = project.create_workflow(name="30-seconds-workflow", 
                                   blocks=['sobloo-s2-l1c-aoiclipped', 'sharpening'])
```
instead of

```
workflow = project.create_workflow(name="30-seconds-workflow")
workflow.add_workflow_tasks(input_tasks=input_tasks)
```

`workflow.add_workflow_tasks()` would be less important and essentially only used under the hood or to alter an existing workflow object.

Function/argument naming should be overhauled at some point (with save deprecation), e.g. add_workflow_tasks is unneccessarily long.

### DIFF
--- a/examples/guides/detailed-example.ipynb
+++ b/examples/guides/detailed-example.ipynb
@@ -76,46 +76,26 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Create a new, empty workflow.\n",
-    "\n",
+    "# Either create a workflow directly filled with tasks\n",
+    "workflow = project.create_workflow(name=\"30-seconds-workflow\",\n",
+    "                                   blocks=['sobloo-s2-l1c-aoiclipped', 'sharpening'],\n",
+    "                                   use_existing=True)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Or Create a new, empty workflow an fill/edit the tasks manually\n",
     "workflow = project.create_workflow(name=\"30-seconds-workflow\", use_existing=False)\n",
-    "workflow"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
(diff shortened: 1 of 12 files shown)


examples/guides/detailed-example.ipynb
```diff
@@ -100,22 +88,14 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Alternative: Add workflow tasks - complex version, gives you more control about the block connections.\n",
+    "# Or Create a new, empty workflow an fill/edit the tasks manually\n",
```
thaisbendixen42, line 91: Or **c**reate a new, empty workflow an**d** fill/edit the tasks manually\n

examples/radar_processing_1.ipynb
```diff
@@ -107,8 +116,19 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Define aoi and input parameters\n",
+    "\n",
+    "The S1 GRD block always delivers the complete images as they are delivered in SAFE format which cannot be clipped.\n",
+    "The image can be clipped though by the SNAP polarimetric processing blocks. For this it is necessary to supply the same geometry (in this case a bbox) to that block as well. \n",
```
thaisbendixen42, line 130: Style suggestion: However, the image can be clipped using the SNAP polarimetric processing **block**.

markusuwe (commented):
It is a nice improvement, but I have my doubts if we should walk down this road even further. This is adding another convenience function on top of the API rather than reflecting API changes 1:1 in the SDK. It does give a better user experience, but the SDK is moving away from the raw API more and more.

chrieke:
Closing this stale PR. Idea can maybe be revisited when finishing the templates.

PR #133, opened 2020-09-11
Remove assertion

nekhtiari:
Remove assertion, fix try except for catalog.py plus adding tests.

### DIFF
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -134,6 +134,47 @@ def test_download_quicklook(catalog_mock):
         assert Path(out_paths[0]).suffix == ".jpg"
 
 
+def test_download_no_quicklook(catalog_mock):
+    sel_id = "dfc54412-8b9c-45a3-b46a-dd030a47c2f3"
+    with tempfile.TemporaryDirectory() as tempdir:
+        provider = "sobloo-image"
+        with requests_mock.Mocker() as m:
+            url = f"{catalog_mock.auth._endpoint()}/catalog/{provider}/image/{sel_id}/quicklook"
+            m.get(url, status_code=404)
+
+            out_paths = catalog_mock.download_quicklooks(
+                image_ids=[sel_id], sensor="sentinel5p", output_directory=tempdir
+            )
+
+        assert len(out_paths) == 0
+
+
+def test_download_1_quicklook_1_no_quicklook(catalog_mock):
+    sel_id_no = "dfc54412-8b9c-45a3-b46a-dd030a47c2f3"
+    sel_id = "6dffb8be-c2ab-46e3-9c1c-6958a54e4527"
+    with tempfile.TemporaryDirectory() as tempdir:
+        provider = "sobloo-image"
+        with requests_mock.Mocker() as m:
+            url_no = f"{catalog_mock.auth._endpoint()}/catalog/{provider}/image/{sel_id_no}/quicklook"
+            m.get(url_no, status_code=404)
+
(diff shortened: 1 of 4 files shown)


tests/test_catalog.py
```diff
@@ -134,6 +134,47 @@ def test_download_quicklook(catalog_mock):
...
+            quicklook_file = (
+                Path(__file__).resolve().parent / "mock_data/a_quicklook.png"
+            )
+            m.get(url, content=open(quicklook_file, "rb").read())
+
+            out_paths = catalog_mock.download_quicklooks(
+                image_ids=[sel_id, sel_id_no],
+                sensor="sentinel5p",
+                output_directory=tempdir,
+            )
```
rodrigoalmeida94, line 171: Nice!!!

rodrigoalmeida94 (approved):
LGTM

up42/catalog.py
```diff
@@ -263,23 +262,21 @@ def download_quicklooks(
...
-
-            with open(out_path, "wb") as dst:
-                for chunk in response:
-                    dst.write(chunk)
+                out_path = output_directory / f"quicklook_{image_id}.jpg"
+                out_paths.append(str(out_path))
+                with open(out_path, "wb") as dst:
+                    for chunk in response:
+                        dst.write(chunk)
+            except ValueError:
```
chrieke, line 276: Which line from the above try: block would trigger the ValueError? Probably only that line should be in the try block to avoid other unforseen Valueerrors.

up42/catalog.py
```diff
@@ -263,23 +262,21 @@ def download_quicklooks(
...
+                logger.warning(
+                    f"{image_id} does not have quicklook available. Skipping ..."
```
chrieke, line 278:
```suggestion
                    f"Image with id {image_id} does not have quicklook available. Skipping ..."
```

tests/test_catalog.py
chrieke: +1

chrieke (commented):
LGTM!

up42/catalog.py
nekhtiari: it is line 268, where else do you think the try mechanism we would get valueerror? SO we thought first we will try to get the quick looks, if they exist, we then write them to the file otherwise we need to skip that iteration of for loop.

chrieke: I see, didn't think about that the raise is gone and this means the whole block needs to be in try/except, ty.

chrieke: Guess best solution to communicate where the valueerror applies & avoid other Valueerrors triggering the message would be like this:

        for image_id in tqdm(image_ids):
            url = f"{self.auth._endpoint()}/catalog/{provider}/image/{image_id}/quicklook"
            try:
                response = self.auth._request(
                    request_type="GET", url=url, return_text=False
                )
            except ValueError:
                logger.warning(
                    f"Image with id {image_id} does not have quicklook available. Skipping ..."
                )
                continue

            out_path = output_directory / f"quicklook_{image_id}.jpg"
            out_paths.append(str(out_path))
            with open(out_path, "wb") as dst:
                for chunk in response:
                dst.write(chunk)

nekhtiari: shit, I merged the branch, if you want I can do it like this, sorry I though you are also fine with the implemetation

chrieke: Your call, if you think an improvement would be good to add.

nekhtiari: yeah, maybe when we want to revisit error handling and throwing exceptions, we can look into this again.

PR #138, opened 2020-09-20
Use correct cutoff time in default parameters

chrieke:
- Uses 23:59:59 for the end date default parameters instead of 00:00:00 which lead to cutoff date 1 day too early.
- Misc small fixes in code & docs.


examples/guides/detailed-example.ipynb
```diff
@@ -377,7 +366,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.4"
+   "version": "3.8.3"
```
markusuwe, line 369: Which version does this refer to?

markusuwe (approved):
LGTM
One question about what I presume is a python version. Would advice to be careful to move to 3.8, might make integration with some standard libraries difficult.

up42/workflow.py
```diff
@@ -327,7 +327,7 @@ def construct_parameters(
                 input_parameters[data_block_name]["limit"] = len(scene_ids)
                 input_parameters[data_block_name].pop("time")
             elif start_date is not None and end_date is not None:
-                datetime = f"{start_date}T00:00:00Z/{end_date}T00:00:00Z"
+                datetime = f"{start_date}T00:00:00Z/{end_date}T23:59:59Z"
                 input_parameters[data_block_name]["time"] = datetime
```
rodrigoalmeida94, line 331: I guess there should be a tests specifically for this no? I see you changed the params but not sure if you input the same the result would be different.

rodrigoalmeida94 (approved):
LGTM! Just one note on tests

examples/guides/detailed-example.ipynb
chrieke: Yeah is 3.8, SDK supports and is regularly tested with 3.8 for some time though. Think we should also move soon to 3.8 in blocks, put it on agenda of next guild meeting.

markusuwe: No need to put it on the agenda. Next meeting will be Tech debt backlog grooming anyway and there already is a ticket for that.

up42/workflow.py
```diff
@@ -327,7 +327,7 @@ def construct_parameters(
                 input_parameters[data_block_name]["limit"] = len(scene_ids)
                 input_parameters[data_block_name].pop("time")
             elif start_date is not None and end_date is not None:
-                datetime = f"{start_date}T00:00:00Z/{end_date}T00:00:00Z"
+                datetime = f"{start_date}T00:00:00Z/{end_date}T23:59:59Z"
```
marooned20, line 330: Although there is no `datetime` module import but minor suggestion would be to name the variable different from module name
```suggestion
                in_datetime = f"{start_date}T00:00:00Z/{end_date}T23:59:59Z"
```

marooned20 (approved):
LGTM

up42/workflow.py
chrieke: renamed

up42/workflow.py
chrieke: Hm guess more of an api behaviour than something that we need to test in the SDK. Just tried in Postman and can confirm that `"2019-01-14T00:00:00Z/2019-01-15T23:59:59Z"` can find a scene on the 15th, but `"2019-01-14T00:00:00Z/2019-01-15T00:00:00Z"` does not. Will change the default dates in the postman examples.

PR #139, opened 2020-09-23
Download jobtasks to subfolder - up4865

chrieke:
When using the default output dir, Jobtasks were downloaded to the jobs folder. When downloading multiple jobtasks this lead to overwritten files. Now downloading to cwd/project_id/job_id/jobtask_id.

### DIFF
--- a/up42/jobtask.py
+++ b/up42/jobtask.py
@@ -99,7 +99,8 @@ class JobTask(Tools):
 
         if output_directory is None:
             output_directory = (
-                Path.cwd() / f"project_{self.auth.project_id}" / f"job_{self.job_id}"
+                Path.cwd()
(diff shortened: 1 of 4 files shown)


up42/jobtask.py
```diff
@@ -99,7 +99,8 @@ def download_results(
 
         if output_directory is None:
             output_directory = (
-                Path.cwd() / f"project_{self.auth.project_id}" / f"job_{self.job_id}"
+                Path.cwd()
+                / f"project_{self.auth.project_id}/job_{self.job_id}/jobtask_{self.jobtask_id}"
```
rodrigoalmeida94, line 103: Nice!

rodrigoalmeida94 (approved):
LGTM

PR #141, opened 2020-09-24
Error messages up4728

chrieke:
Removes explicit error handling and passes through error message. 

HTTPError or ValueError as the default error?

image

### DIFF
--- a/up42/auth.py
+++ b/up42/auth.py
@@ -31,7 +31,7 @@ class Auth(Tools):
         Info:
             Authentication is possible via the credentials of a specific project (project_id &
             project_api_key). To get your **project id** and **project api key**, follow
-            the instructions in the docs installation chapter.
+            the instructions in the docs authentication chapter.
 
         Args:
             cfg_file: File path to the cfg.json with {project_id: "...", project_api_key: "..."}.
--- a/up42/auth.py
+++ b/up42/auth.py
@@ -111,10 +111,10 @@ class Auth(Tools):
     def _get_token(self):
         try:
             self._get_token_project()
-        except requests.exceptions.HTTPError:
+        except HTTPError as err:
             raise ValueError(
-                "Authentication was not successful, check the provided project keys."
-            )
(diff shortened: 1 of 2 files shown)


rodrigoalmeida94:
> Removes explicit error handling and passes through error message.
> 
> HTTPError or ValueError as the default error?
> 
> image

I think in the Ideal world we would define our own UP42SDKError class that would be specific for this context. For now we could just raise RequestsException. Using ValueError can cause some confusion if you want to do try: except statements.

up42/auth.py
```diff
@@ -111,10 +111,10 @@ def _endpoint(self) -> str:
     def _get_token(self):
         try:
             self._get_token_project()
-        except requests.exceptions.HTTPError:
+        except requests.exceptions.HTTPError as err:
             raise ValueError(
```
rodrigoalmeida94, line 115: What does this `from err` do? This is a new notation for me :)

up42/auth.py
```diff
@@ -224,18 +224,13 @@ def _request(
...
-                )
-            elif response.status_code == 404:
-                raise ValueError("Product not found!")
+        try:
+            response.raise_for_status()
+        except requests.exceptions.RequestException as err:  # Base error class
+            err_message = json.loads(response.text)["error"]
+            err_message = f"{err_message['code']} Error - {err_message['message']}!"
+            logger.error(err_message)
+            raise requests.exceptions.HTTPError(err_message) from err
```
rodrigoalmeida94, line 233: Nice!

rodrigoalmeida94 (approved):
LGTM

marooned20 (approved):
LGTM!

up42/auth.py
chrieke: Similar to the stacktrace in the screenshot example, is a bit more explicit that this custom Exception is a direct consequence of the previous exception. Is a detail but fits well here I think.

PR #143, opened 2020-09-28
Less verbose repr and logger

chrieke:
- Simplifies the object __repr__ to only include relevant information. The verbose output also blew up the logger messages and made it difficult to work with Jobcollection/lists of objects (Also suggested in feedback). `.info` for the full metadata as json, will also be made clearer in the docs.

- `get_workflow_tasks(basic=True)` now returns task_name: block_version instead of task_name: task_id (which is never used anyway).
- Adds workflow_tasks property
- Adds workflow tasks to workflow __repr__

- Simplifies the logger message format & truncates in case of extremely log messages (e.g. large collections).

Before/After:
image
image

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -200,7 +200,7 @@ def jobtask_mock(auth_mock):
             f"{auth_mock._endpoint()}/projects/{auth_mock.project_id}/jobs/{job_id}"
             f"/tasks/"
         )
-        m.get(url=url_jobtask_info, json={"data": {"xyz": 789}, "error": {}})
+        m.get(url=url_jobtask_info, json={"data": [{"xyz": 789}], "error": {}})
 
         jobtask = JobTask(
             auth=auth_mock,
--- a/tests/test_workflow.py
+++ b/tests/test_workflow.py
@@ -32,6 +32,7 @@ json_workflow_tasks = {
         {
             "id": "c0d04ec3-98d7-4183-902f-5bcb2a176d89",
             "name": "sobloo-s2-l1c-aoiclipped:1",
+            "blockVersionTag": "2.2.2",
             "block": {
                 "name": "sobloo-s2-l1c-aoiclipped",
                 "parameters": {
@@ -48,6 +49,7 @@ json_workflow_tasks = {
         {
             "id": "af626c54-156e-4f13-a743-55efd27de533",
             "name": "tiling:1",
+            "blockVersionTag": "1.0.0",
             "block": {
                 "name": "tiling",
                 "parameters": {
@@ -156,6 +158,7 @@ def test_get_workflow_tasks_normal_and_basic(workflow_mock):
     assert tasks[0] == {
         "id": "c0d04ec3-98d7-4183-902f-5bcb2a176d89",
         "name": "sobloo-s2-l1c-aoiclipped:1",
+        "blockVersionTag": "2.2.2",
         "block": {
             "name": "sobloo-s2-l1c-aoiclipped",
             "parameters": {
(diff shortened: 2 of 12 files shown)


thaisbendixen42 (approved):
Output looks a lot better!!

tests/fixtures.py
```diff
@@ -206,7 +206,7 @@ def jobtask_mock(auth_mock):
             f"{auth_mock._endpoint()}/projects/{auth_mock.project_id}/jobs/{job_id}"
             f"/tasks/"
         )
-        m.get(url=url_jobtask_info, json={"data": {"xyz": 789}, "error": {}})
+        m.get(url=url_jobtask_info, json={"data": [{"xyz": 789}], "error": {}})
```
rodrigoalmeida94, line 209: Is data now a list? Did this change in the API?

tests/test_workflow.py
```diff
@@ -156,6 +158,7 @@ def test_get_workflow_tasks_normal_and_basic(workflow_mock):
     assert tasks[0] == {
         "id": "c0d04ec3-98d7-4183-902f-5bcb2a176d89",
         "name": "sobloo-s2-l1c-aoiclipped:1",
+        "blockVersionTag": "2.2.2",
```
rodrigoalmeida94, line 161: Cool! Does the API now return the version tag too? Or did it always do that?

up42/__init__.py
```diff
@@ -57,30 +61,45 @@ def initialize_workflow(workflow_id) -> "Workflow":
...
+    jobtask = JobTask(
         auth=_auth,
         jobtask_id=jobtask_id,
         job_id=job_id,
         project_id=str(_auth.project_id),
     )
+    try:
+        logger.info(f"Initialized {jobtask}")
+    except AttributeError:
+        pass
```
rodrigoalmeida94, line 101: Why do we need the try except?

up42/workflow.py
```diff
@@ -95,7 +108,7 @@ def get_workflow_tasks(self, basic: bool = False) -> Union[List, Dict]:
         logger.info(f"Got {len(tasks)} tasks/blocks in workflow {self.workflow_id}.")
 
         if basic:
-            return {task["name"]: task["id"] for task in tasks}
+            return {task["name"]: task["blockVersionTag"] for task in tasks}
```
rodrigoalmeida94, line 111: Nice! Can we construct a new workflow from this dictionary?

rodrigoalmeida94 (approved):
LGTM! Some questions

tests/test_workflow.py
chrieke: Always returned it, now just added as default output to `basic=True` as well so needed to add to mock test.

up42/workflow.py
chrieke: How do you mean, as a functionality to clone a workflow?

rodrigoalmeida94: Yes - intuitively I would say it this returns a dict, I should be able to use this dict to add tasks to a workflow.

chrieke: Hm yes, we could have this format as the default option to add blocks with a specific version, would be nice:

So simple format to get the newest block version:
`workflow.add_workflow_tasks(["blockname1", "blockname2"])`

or to specify the block version:
`workflow.add_workflow_tasks({"blockname1":"version1", "blockname2":"version2"})`

The only problem is that the output of `get_workflow_tasks(basic=True)` and `workflow_tasks` also contains the additional `:1` as the workflow task structure indication. Guess we either remove that here or check/handle it in `add_workflow_tasks`.

rodrigoalmeida94: Hm ok, seems a bit more complicated, maybe we do it some other time, since we want to deploy.

tests/fixtures.py
chrieke: Yes jobtasks data is always a list of dictionaries. mock test wasn't correct, added a list assertion now.

up42/__init__.py
chrieke: Not required, wrong thought, thanks!

PR #145, opened 2020-09-29
Add selection of mode when getting jobs

rodrigoalmeida94:
Adds a method to Tools to filter by "mode" (DEFAULT or DRY_RUN). Implements flags in Workflow.get_jobs and Project.get_jobs to filter by mode.

### DIFF
--- a/up42/tools.py
+++ b/up42/tools.py
@@ -447,3 +447,32 @@ class Tools:
         )
         logger.info("The manifest is valid.")
         return response_json["data"]
+
+    @staticmethod
+    def filter_jobs_on_mode(
+        jobs_json: List[dict], test_jobs: bool = True, real_jobs: bool = True
+    ) -> List[dict]:
+        """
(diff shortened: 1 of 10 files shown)


up42/tools.py
```diff
@@ -447,3 +447,32 @@ def validate_manifest(self, path_or_json: Union[str, Path, Dict]) -> Dict:
         )
         logger.info("The manifest is valid.")
         return response_json["data"]
+
+    @staticmethod
+    def filter_jobs_on_mode(
```
chrieke, line 452: Would probably be better in utils.

chrieke (commented):
LGTM, nice! Would probably be better placed into utils module.

up42/tools.py
rodrigoalmeida94: Moved.

chrieke (approved):

PR #148, opened 2020-10-08
Info property up4885

chrieke:
- Adds `info` property.
- Adds `status` property.
- Adds deprecation warnings to `get_info` & `get_status`

Future:
- Keeping the extra `_info` attribute for now, first need to simplify the mock tests with `requests_mock` so that we can use mocks in the fixture.
- Add all suitable "get" functions as class properties, e.g. "project.settings", "workflow.workflow_tasks" etc. for better usability.

Trying to keep the attributes/property section as clean as possible:
image

Gotta love mkdocs, linkthrough:

image

### DIFF
--- a/up42/job.py
+++ b/up42/job.py
@@ -71,11 +71,11 @@ class Job(Tools):
         logger.info(f"Job is {status}")
         return status
 
+    @deprecation("get_status", "job.status")
     def get_status(self) -> str:
         """
         `get_status` will be deprecated in release 0.13, use [status attribute](job.md#up42.job.Job.status) instead.
         """
-        deprecation("get_status", "job.status")
         return self.status
 
     @property
--- a/up42/job.py
+++ b/up42/job.py
@@ -71,6 +71,13 @@ class Job(Tools):
         logger.info(f"Job is {status}")
         return status
 
+    def get_status(self) -> str:
+        """
+        `get_status` will be deprecated in release 0.13, use [status attribute](job.md#up42.job.Job.status) instead.
+        """
+        deprecation("get_status", "job.status")
+        return self.status
+
     @property
     def is_succeeded(self) -> bool:
         """
--- a/up42/job.py
+++ b/up42/job.py
@@ -59,7 +59,7 @@ class Job(Tools):
         url = f"{self.auth._endpoint()}/projects/{self.project_id}/jobs/{self.job_id}"
         response_json = self.auth._request(request_type="GET", url=url)
         self._info = response_json["data"]
-        return self._info
+        return response_json["data"]
 
     @property
     def status(self) -> str:
--- a/up42/job.py
+++ b/up42/job.py
@@ -51,8 +51,13 @@ class Job(Tools):
         )
 
     @property
-    def info(self):
-        """Gets metadata info from an existing Job"""
+    def info(self) -> Dict:
+        """
+        Gets the job metadata information.
+
+        Returns:
+            A dictionary with the job metadata information.
+        """
         url = f"{self.auth._endpoint()}/projects/{self.project_id}/jobs/{self.job_id}"
         response_json = self.auth._request(request_type="GET", url=url)
         self._info = response_json["data"]
@@ -61,7 +66,7 @@ class Job(Tools):
     @property
     def status(self) -> str:
         """
-        Gets the job status.
+        Gets the job progress status.
 
         Returns:
             The job status, one of "SUCCEEDED", "NOT STARTED", "PENDING", "RUNNING",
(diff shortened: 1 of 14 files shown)


up42/job.py
```diff
@@ -42,41 +43,51 @@ def __init__(
...
         url = f"{self.auth._endpoint()}/projects/{self.project_id}/jobs/{self.job_id}"
         response_json = self.auth._request(request_type="GET", url=url)
-        self.info = response_json["data"]
-        return self.info
+        self._info = response_json["data"]
+        return self._info
 
-    def get_status(self) -> str:
+    @property
+    def status(self) -> str:
```
markusuwe, line 65: Why the renaming? In case of the SDK, renaming always leads to breaking code which uses this functionality.

up42/job.py
```diff
@@ -42,41 +43,51 @@ def __init__(
...
+            `RUNNING`, `CANCELLED`, `CANCELLING`, `FAILED`, `ERROR`.
         """
-        info = self._get_info()
-        status = info["status"]
+        status = self.info["status"]
         logger.info(f"Job is {status}")
         return status
 
+    @deprecation("get_status", "job.status")
+    def get_status(self) -> str:
```
markusuwe, line 75: I see, you even thought of deprecation.

markusuwe (approved):
LGTM
Needs more reviews. The deprecation mechanism is a new pattern that all contributors must adhere to and therefore need to learn about it in the first place.

up42/job.py
chrieke: Good question, feels very "natural" to use to have these as object properties, even though does a request under the hood. SDK requires an internet connection anyway. Rodrigo added some status queries as class properties and we really liked it. Initial motivation was to consolodiate the naming with the differently called jobcollection functions, now combined with the property change.

nekhtiari (approved):
Nice!

up42/job.py
markusuwe: Merci

up42/job.py
```diff
@@ -42,41 +43,51 @@ def __init__(
         self.results = None
         self.order_ids = order_ids
         if self.auth.get_info:
-            self.info = self._get_info()
+            self._info = self.info
```
rodrigoalmeida94, line 46: Do we need to define _info as an instance variable? Would users interact with this variable directly? If not would it not make sense to remove it entirely?

up42/job.py
```diff
@@ -42,41 +43,51 @@ def __init__(
...
+        Gets the job progress status. One of `SUCCEEDED`, `NOT STARTED`, `PENDING`,
+            `RUNNING`, `CANCELLED`, `CANCELLING`, `FAILED`, `ERROR`.
         """
-        info = self._get_info()
-        status = info["status"]
+        status = self.info["status"]
         logger.info(f"Job is {status}")
         return status
 
+    @deprecation("get_status", "job.status")
```
rodrigoalmeida94, line 74: Oh cool! What does this do exactly?

up42/jobtask.py
```diff
@@ -32,23 +32,26 @@ def __init__(
         self.quicklooks = None
         self.results = None
         if self.auth.get_info:
-            self.info = self._get_info()
+            self._info = self.info
```
rodrigoalmeida94, line 35: Same here

up42/utils.py
```diff
@@ -46,6 +48,37 @@ def get_logger(name, level=logging.INFO):
...
+            message = (
+                f"`{function_name}` will be deprecated in version {version}, "
+                f"use `{replacement_name}` instead! {extra_message}"
+            )
+            warnings.warn(message, DeprecationWarning, stacklevel=2)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return actual_decorator
```
rodrigoalmeida94, line 79: Oh beautiful!! Gotta love the decorators ❤️

up42/workflow.py
```diff
@@ -35,23 +35,26 @@ def __init__(self, auth: Auth, project_id: str, workflow_id: str):
...
+        """
         url = (
             f"{self.auth._endpoint()}/projects/{self.project_id}/workflows/"
             f"{self.workflow_id}"
         )
         response_json = self.auth._request(request_type="GET", url=url)
-        self.info = response_json["data"]
-        return self.info
+        self._info = response_json["data"]
+        return self._info
```
rodrigoalmeida94, line 57: I think here we could just return `response_json["data"]`

rodrigoalmeida94 (approved):
LGTM! Just some question around having the instance variable

up42/job.py
chrieke: Users would not interact with it, just so that the information is stored in the object and can be used internally. I want to ditch it as discussed, but would currently require too many code changes with the tests, easier when we first establish using mocks in the fixtures, will create a task for both things. Using `_info` right now basically establishes the behaviour of how it was before, not ideal, but better to do it in steps and you still get the `info` property. Should work for now (?).

up42/job.py
chrieke: If you use a decorated function just gives a warnings.warn Deprecation message to the user.

up42/jobtask.py
chrieke: see above

up42/workflow.py
chrieke: thanks

PR #150, opened 2020-10-13
Code reference improvements

chrieke:
Last PR before new version.

- Add object initialization at top of page for each object to code reference
- Add up42 base import chapter to code reference (currenty only contains the object initializations, to be extended with get_blocks() etc when inheritance is clearer).
- Remove `order_ids` from `up42.initialize_job`, makes no sense, only at Job creation.

### DIFF
--- a/docs/reference/job.md
+++ b/docs/reference/job.md
@@ -1,7 +1,20 @@
-# Job class
+# Job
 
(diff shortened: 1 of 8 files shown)


rodrigoalmeida94 (approved):
LGTM

thaisbendixen42 (approved):
LGTM

PR #152, opened 2020-10-13
unpin geopandas version

chrieke:
Umpin geopandas version from 0.7.0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ click
 tqdm
 geojson
 rasterio
-geopandas==0.7.0
+geopandas
 folium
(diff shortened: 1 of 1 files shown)


thaisbendixen42 (approved):

rodrigoalmeida94 (approved):
LGTM! Should we unpin in blocks as well? We pinned it because of a bug rigth?

markusuwe:
> LGTM! Should we unpin in blocks as well? We pinned it because of a bug rigth?

Sounds like a good idea.

markusuwe (approved):
LGTM

PR #158, opened 2020-10-28
Deprecate tools

chrieke:


docs/reference/catalog.md
```diff
@@ -1,8 +1,16 @@
...
 
 
-::: up42.catalog.Catalog
\ No newline at end of file
+::: up42.catalog.Catalog
+    rendering:
+        show_root_toc_entry: False
+    selection:
+        inherited_members: True
+        members: ["construct_parameters", "search", "download_quicklooks", "plot_quicklooks", "map_quicklooks"]
```
chrieke, line 15: Less then ideal, but currently no way to exclude specific members, only via not including them. Feature request in mkdocstrings: https://github.com/pawamoy/mkdocstrings/issues/157

up42/__init__.py
```diff
@@ -131,20 +131,17 @@ def validate_manifest(path_or_json: Union[str, Path, Dict]) -> Dict:
 def read_vector_file(
     filename: str = "aoi.geojson", as_dataframe: bool = False
 ) -> FeatureCollection:
-    tools = Tools(auth=_auth)
-    return tools.read_vector_file(filename, as_dataframe)
+    return Tools().read_vector_file(filename, as_dataframe)
```
rodrigoalmeida94, line 134: Why are we instantiating the Tools object? If read_vector_file is a static method we can just do `Tools.read_vector_file()` and avoid creating the object.

up42/tools.py
```diff
@@ -39,6 +40,25 @@ def __init__(self, auth=None):
         self.quicklooks = None
         self.results = None
 
+    def _deprecate_tools(self, function_name: str):
+        # TODO: When finallay deprecating this, move the functions that don't require
```
rodrigoalmeida94, line 44: Finally

up42/tools.py
```diff
@@ -39,6 +40,25 @@ def __init__(self, auth=None):
...
+            "Job",
+            "JobTask",
+            "JobCollection",
+            "Catalog",
+        ]:
+            message = (
+                f"Use of {object_name}.{function_name} will be deprecated in "
+                f"version 0.14.0, use up42.{function_name} instead!"
+            )
+            warnings.warn(message, DeprecationWarning, stacklevel=3)
```
rodrigoalmeida94, line 60: Nice solution!

rodrigoalmeida94 (approved):
LGTM

up42/__init__.py
chrieke: Needs the self for using `self._deprecate_tools` in the function. Which needs to be within the class as it requires access to `self.__class__.__name__`. Added as reminder to the todo to remove the instantiation when finally deprecated, thanks.

PR #167, opened 2020-11-16
Add exponential backoff up4980

rodrigoalmeida94:
With newly introduced rate limiting in the UP42 API, this PR handles retrying in cases where the rate limit is exceed by using a exponential backoff strategy with jitter (see [tenacity docs](https://tenacity.readthedocs.io/en/latest/api.html#tenacity.wait.wait_random_exponential)).

### DIFF
--- a/up42/auth.py
+++ b/up42/auth.py
@@ -10,13 +10,33 @@ import requests.exceptions
 from tenacity import (
     Retrying,
     wait_fixed,
+    wait_random_exponential,
     stop_after_attempt,
+    retry_if_exception,
     retry_if_exception_type,
+    retry,
 )
 
 from up42.utils import get_logger
 
 logger = get_logger(__name__)
+
+
+class retry_if_429_error(retry_if_exception):
+    """
+    Adapted from https://github.com/alexwlchan/handling-http-429-with-tenacity
+    Retry strategy that retries if the exception is an ``HTTPError`` with
+    a 429 status code.
+    """
+
+    def __init__(self):
+        def is_http_429_error(exception):
+            return (
+                isinstance(exception, requests.exceptions.HTTPError)
+                and exception.response.status_code == 429
+            )
+
+        super().__init__(predicate=is_http_429_error)
 
 
 class Auth:
@@ -152,6 +172,11 @@ class Auth:
         return headers
 
     # pylint: disable=dangerous-default-value
+    @retry(
+        retry=retry_if_429_error(),
+        wait=wait_random_exponential(multiplier=0.5, max=180),
+        reraise=True,
+    )
     def _request_helper(
         self, request_type: str, url: str, data: Dict = {}, querystring: Dict = {}
     ) -> requests.Response:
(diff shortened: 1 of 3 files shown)


up42/auth.py
```diff
@@ -210,26 +236,25 @@ def _request(
...
+                    reraise=True,
+                )
+                response = retryer(
+                    self._request_helper, request_type, url, data, querystring
+                )
+            else:
+                response = self._request_helper(request_type, url, data, querystring)  # type: ignore
         except requests.exceptions.RequestException as err:  # Base error class
-            err_message = json.loads(response.text)["error"]
+            err_message = json.loads(err.response.text)["error"]
```
chrieke, line 257: o wow, good catch!

up42/auth.py
```diff
@@ -10,15 +10,35 @@
...
 )
 
 from up42.utils import get_logger
 
 logger = get_logger(__name__)
 
 
+class retry_if_429_error(retry_if_exception):
+    """
+    Adapted from https://github.com/alexwlchan/handling-http-429-with-tenacity
```
chrieke, line 27: nice!

chrieke (approved):
LGTM, nice solution. We can maybe later combine the `retry_if_429_error` retry reason with the other retry mechanism when that is activated again.

up42/auth.py
```diff
@@ -10,15 +10,35 @@
...
+    """
+
+    def __init__(self):
+        def is_http_429_error(exception):
+            return (
+                isinstance(exception, requests.exceptions.HTTPError)
+                and exception.response.status_code == 429
+            )
+
+        super().__init__(predicate=is_http_429_error)
```
nekhtiari, line 39: What does this predicate do here?

nekhtiari (approved):
Nice

up42/auth.py
rodrigoalmeida94: the predicate comes from the super call (retry_if_exception from tenacity). It's a method that return True of False, in essence.

up42/auth.py
```diff
@@ -210,26 +236,25 @@ def _request(
...
-                after=self._get_token(),
-            )
-            response = retryer(
-                self._request_helper, request_type, url, data, querystring
-            )
-        else:
-            response = self._request_helper(request_type, url, data, querystring)  # type: ignore
-
         try:
-            response.raise_for_status()
```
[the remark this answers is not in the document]
chrieke: Yeah the current retry mechanism is not well implemented, was deactivated with `stop_after_attempt(1)`. The self.retry flag is important for some tests where you don't want to see retry behaviour. I think this can be refactored and maybe just combined into your retry mechanism with multiple triggers (or maybe it needs to stay separate if we think the rate limiting should have longer wait times etc). Anyway, would say we merge this and create a techdebt ticket to refactor the retry mechanism + self.retry flag as a whole. Or should look into it right now?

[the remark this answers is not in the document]
rodrigoalmeida94: Sounds good! I was just afraid maybe this messes some things up. If your comfortable I will then merge and we tackle retrying later.

PR #171, opened 2020-11-30
Add price estimation endpoint in SDK

nekhtiari:


tests/fixtures.py
```diff
@@ -36,48 +37,63 @@
...
-                        "description": "Value representing..",
+                        "default": "None",
+                        "required": "false",
+                        "description": "Value representing ...",
                     },
                     "tile_width": {
                         "type": "number",
                         "default": 768,
-                        "required": True,
+                        "required": "true",
```
rodrigoalmeida94, line 86: Why does this need to change? In general I think when you define a dictionary, it should have boolean values set to `True` and `False` (they get converted from json `true` and `false) and the same with `None` (gets converted from json `null`).

up42/workflow.py
```diff
@@ -603,9 +631,10 @@ def _helper_run_parallel_jobs(
 
     def test_job(
         self,
-        input_parameters: Union[Dict, str, Path] = None,
+        input_parameters: Union[Dict, str, Path],
```
rodrigoalmeida94, line 634: Why does this need to change? I see there is some handling of None in the previous function, maybe it would be good to add the same for estimation?

rodrigoalmeida94 (approved):
Cool! Two small notes

tests/fixtures.py
nekhtiari: oh right, fixed it

up42/workflow.py
nekhtiari: Thanks, will revert this, I had some pylint issue and when I fixed it I forget to revert this back.

up42/workflow.py
```diff
@@ -442,6 +444,32 @@ def construct_parameters_parallel(
...
+        full_input_tasks = self._construct_full_workflow_tasks_dict(input_tasks)
+        for index, _ in enumerate(workflow_tasks):
+            full_input_tasks[index]["blockVersionTag"] = workflow_tasks[index][
+                "blockVersionTag"
+            ]
+
+        estimation = Estimation(
+            self.auth, self.project_id, input_parameters, full_input_tasks
+        ).estimate_price()
+        return estimation
```
chrieke, line 471: Logger message for "one glimpse info" would be cool, e.g.:

```
        min_credits, max_credits, min_duration, max_duration = [], [], [], []
        for e in estimation.values():
            min_credits.append(e["blockConsumption"]["credit"]["min"])
            max_credits.append(e["blockConsumption"]["credit"]["max"])
            min_credits.append(e["machineConsumption"]["credit"]["min"])
            max_credits.append(e["machineConsumption"]["credit"]["max"])

            min_duration.append(e["machineConsumption"]["duration"]["min"])
            max_duration.append(e["machineConsumption"]["duration"]["max"])

        logger.info(f"Estimated: {sum(min_credits)}-{sum(max_credits)} Credits, "
                    f"Duration: {int(sum(min_duration)/60)}-{int(sum(max_duration)/60)} min.")
```

up42/workflow.py
```diff
@@ -442,6 +444,32 @@ def construct_parameters_parallel(
 
         return result_params
 
+    def estimate_job(self, input_parameters: Union[Dict, str, Path] = None) -> Dict:
+        """
+        A function to get the job estimation.
```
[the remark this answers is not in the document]
chrieke: Not sure about the range, need to check with Garima about official communication. Would be good to give here since showing up in code reference.

chrieke (commented):
Nice with the class! Some suggestions for log message.

up42/workflow.py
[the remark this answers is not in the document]
nekhtiari: I also do not know

markusuwe (approved):
LGTM
How about adding this right away to one of the examples?

nekhtiari:
> LGTM
> How about adding this right away to one of the examples?

I can add it to 30 second one?

PR #172, opened 2020-12-03
Update streaming examples

markusuwe:
Ship detection no uses download data, aircraft detection only displays plane bounding boxes without image.


chrieke:
Ship detection LGTM.

Aircraft detection would update the visualization to this, map context looks better than the white plots & more compact code possible. We could also update .map_results() to plot the image + the detections in the interactive map, task is in the backlog. Would be cool for the webinar.

image

marooned20 (approved):
LGTM!
I agree with @chrieke about adding a map extent instead of blank canvas. Would enhance the messaging indeed.
Minor suggestion would be to add `[Ship Identification on UP42 MArketplace](https://marketplace.up42.com/block/54217695-73f4-4528-a575-a429e9af6568)` in the notebook instead of plain link!

markusuwe:
> LGTM!
> I agree with @chrieke about adding a map extent instead of blank canvas. Would enhance the messaging indeed.
> Minor suggestion would be to add `[Ship Identification on UP42 MArketplace](https://marketplace.up42.com/block/54217695-73f4-4528-a575-a429e9af6568)` in the notebook instead of plain link!

Good point, thanks.

markusuwe:
> Ship detection LGTM.
> 
> Aircraft detection would update the visualization to this if okay with you, map context looks better than the white plots & more compact code possible. We could also update .map_results() to plot the image + the detections in the interactive map, task is in the backlog. Could be cool for the webinar.
> 
> image

That's a good suggestion in this particular case. FYI, the aircraft detection will not show up in the webinar, I am only doing this update right now because of the streaming data license changes.

PR #174, opened 2020-12-03
Plotting bands

chrieke:
Enables band selection in `plot_results` and `map_results`. Also required for new default examples.

x           |  y |  z
:-------------------------:|:-------------------------:|:-------------------------:
image |  image | image

### DIFF
--- a/up42/viztools.py
+++ b/up42/viztools.py
@@ -53,7 +53,9 @@ class VizTools:
         titles: List[str] = None,
         filepaths: Union[List[Union[str, Path]], Dict] = None,
         plot_file_format: List[str] = [".tif"],
+        **kwargs,
     ) -> None:
+        # pylint: disable=line-too-long
         """
         Plots image data (quicklooks or results)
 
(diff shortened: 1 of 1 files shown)


up42/viztools.py
```diff
@@ -101,11 +103,14 @@ def plot_results(
         else:
             axs = [axs]
 
+        if len(bands) != 3:
+            if len(bands) == 1:
+                bands = bands * 3  # plot as grayband
```
marooned20, line 108: If there is only 1 band, this line duplicates it 3 times? Is this necessary?

markusuwe (approved):
LGTM
Nice that this follows the same syntax that rasterio.plot is using.

thaisbendixen (approved):
LGTM!

up42/viztools.py
chrieke: Yeah that is interesting, not neccessary, but more conveniant to use. If you read just one band, the rasterio show/plot function will apply the default colormap, so the image will be green/blue. Single band should be grayscale imho, so to not have to expose the colormap I thought this would be the best option. Will check if I can also just disable the colormap in case of a single band.

up42/viztools.py
```diff
@@ -47,21 +49,21 @@ def __init__(self):
     def plot_results(
         self,
         figsize: Tuple[int, int] = (14, 8),
-        filepaths: Union[List[Union[str, Path]], Dict] = None,
+        bands: List[int] = [1, 2, 3],
```
rodrigoalmeida94, line 52: Using a Tuple here instead of a list might enable you to remove the pylint disable.

rodrigoalmeida94 (approved):
Cool feature!

up42/viztools.py
chrieke: Using cmap now, more efficient, thanks.

up42/viztools.py
chrieke: I the user wants to give a single band, would need to be explicitly a tuple e.g. with additional comma `bands=(1,)`, probably leads to problems.

PR #176, opened 2020-12-07
New guide examples - up5104

chrieke:
- Adjust 30sec & detailed example (sentinelhub-s2 & landsurface temperature)
- Hide airport monitoring example for now (airport detection is better anyway)


markusuwe (approved):
LGTM

PR #184, opened 2021-01-18
Add ordering up5192 1/3

rodrigoalmeida94:
Adds order feature in SDK. This also includes handlers for storage and assets. 
This PR is 1/3 PRs in order to implement ordering. This includes all the basic object definition and documentation. 

PR 2 will include constructors from other Classes (i.e. Catalog -> Order and Asset -> Workflow). https://github.com/up42/up42-py/pull/186
PR 3 will include a usage example and docs improvements if needed. https://github.com/up42/up42-py/pull/187

Objects:

- [x] Storage
- [x] Asset
- [x] Order

Items:
* [x] Ran test & live-tests
* [x] Removed credentials


tests/test_order.py
```diff
@@ -0,0 +1,243 @@
...
+    asset_mock,
+    asset_live,
+)
+
+
+def test_init(order_mock):
+    assert isinstance(order_mock, Order)
+    assert order_mock.order_id == ORDER_ID
+    assert order_mock.workspace_id == WORKSPACE_ID
+
```
thaisbendixen, line 25: Is this testing something specific? Or just to be extra extra cautious that parameters are getting passed correctly?

rodrigoalmeida94: Essentially this is testing if the __init__ method of the Order class is working as expected - i.e. that the object is initialized in the fixture and that the main attributes are defined correctly.

thaisbendixen (approved):
Really coooooool!!!!! 💥💥💥

up42/storage.py
```diff
@@ -0,0 +1,91 @@
...
+from up42.auth import Auth
+from up42.order import Order
+from up42.asset import Asset
+from up42.utils import get_logger
+from up42.tools import Tools
+
+logger = get_logger(__name__)
+
+
+class Storage(Tools):
```
chrieke, line 14: Can remove Tools from Storage, Assets, Orders. The inheritance from Tools is prepared to get deprecated soon for the other classes, so guess don't even need to introduce here.

up42/storage.py
```diff
@@ -0,0 +1,91 @@
...
+        The Storage class can query all available assets within an UP42 workspace.
+        """
+        self.auth = auth
+        self.workspace_id = auth.workspace_id
+
+    def __repr__(self):
+        env = ", env: dev" if self.auth.env == "dev" else ""
+        return f"Storage(workspace_id: {self.workspace_id}{env})"
+
+    def _paginate(self, url: str) -> List[Dict]:
```
chrieke, line 26: nice!!

up42/storage.py
```diff
@@ -0,0 +1,91 @@
...
+        Returns:
+            List[Dict]: List of all paginated items.
+        """
+        first_pagination_response = self.auth._request(request_type="GET", url=url)
+        output = first_pagination_response["data"]["content"]
+        num_pages = first_pagination_response["data"]["totalPages"]
+        total_items = first_pagination_response["data"]["totalElements"]
+        for page in range(2, num_pages + 1):
+            response_json = self.auth._request(
+                request_type="GET", url=url + f"?page={page}"
```
chrieke, line 42: Please add a small comment that the pages url index is apparently 1-indexed.

up42/storage.py
```diff
@@ -0,0 +1,91 @@
...
+        Helper to fetch list of items in paginated endpoint.
+
+        Args:
+            url (str): The base paginated endpoint.
+
+        Returns:
+            List[Dict]: List of all paginated items.
+        """
+        first_pagination_response = self.auth._request(request_type="GET", url=url)
+        output = first_pagination_response["data"]["content"]
```
chrieke, line 37: Is output already a list here so that you can append to it further below? Looks like it would be a dict here?

up42/order.py
```diff
@@ -0,0 +1,178 @@
...
+        url = f"{auth._endpoint()}/workspaces/{auth.workspace_id}/orders/estimate"
+        payload = {
+            "dataProviderName": data_provider_name,
+            "orderParams": order_params,
+        }
+
+        response_json = auth._request(request_type="POST", url=url, data=payload)
+        estimated_credits: int = response_json["data"]["credits"]  # type: ignore
+        logger.info(
+            f"Order with order parameters {payload} is estimated to cost {estimated_credits}."
```
chrieke, line 138:
```suggestion
            f"Order with order parameters {payload} is estimated to cost {estimated_credits} UP42 credits."
```

up42/order.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        return estimated_credits
+
+    def track_status(self, report_time: int = 120) -> str:
+        """`
```
chrieke, line 143:
```suggestion
        """
```

chrieke (commented):
Awesome!

up42/storage.py
nekhtiari: very cool!

nekhtiari (approved):
Great job!

up42/storage.py
rodrigoalmeida94: Ah cool, missed this completely, thanks.

rodrigoalmeida94: Removing the Tools inheritance gives out mypy errors in all of the classes, seems like we would need to revisit this some other time.

up42/storage.py
rodrigoalmeida94: Pages are indexed from 0 to n pages, so "normal" python indexing. Just realized the range should be from 1 to num_pages +1, thanks!

up42/storage.py
rodrigoalmeida94: Yup, this is already a list - since pagination implies that "content" will include always a list of things.

chrieke (approved):
updates lgtm

up42/order.py
```diff
@@ -0,0 +1,178 @@
...
+from up42.utils import (
+    get_logger,
+)
+
+logger = get_logger(__name__)
+
+DATA_PROVIDERS = ["oneatlas"]
+
+
+class Order(Tools):
```
markusuwe, line 16: Semantically it sounds odd that an Order "is a" tool (which is the meaning of the inheritance relation).

markusuwe (approved):
Very well written code, LGTM.
Not necessarily something to address right now, but something for the future: try to model objects and classes with correct semantics. Inheritance is an "is-a" relationship ("a lion is a cat"). Trying to give proper meaning to classes will help a lot when a framework grows more wrt to readability.

up42/order.py
rodrigoalmeida94: Thanks! You're absolutely right, @chrieke is in the process of deprecating the Tools object, but we encountered some mypy problems once we remove the inheritance. For some reason, once the class has no parent, we get a lot of mypy errors (I believe this is due to the fact that the method `Auth._request` has an undefined type - can be Response, json or text). We want to refactor this method and at that point we could also remove this unnatural inheritances.

PR #185, opened 2021-01-19
Adjust map_results to enable plotting jobcollection

marooned20:
Currently `map_results` only supports a single job and fails if multiple jobs were run (i.e. a jobcollection). This PR enables plotting functionality for jobcollection as well.

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/viztools.py
```diff
@@ -308,29 +322,42 @@ def map_results(
                 "You first need to download the results via job.download_results()!"
             )
 
-        # Add features to map.
-        # Some blocks store vector results in an additional geojson file.
-        # pylint: disable=not-an-iterable
-        json_fp = [fp for fp in self.results if fp.endswith(".geojson")]
```
marooned20, line 314: @chrieke I could not really understand which specific scenario this line handles. Could you give an example for better understanding!?

up42/viztools.py
```diff
@@ -308,29 +322,42 @@ def map_results(
...
+            # Add features to map.
+            # Some blocks store vector results in an additional geojson file.
             # pylint: disable=not-an-iterable
-            json_fp = [fp for fp in self.results if fp.endswith(".json")][0]
+            json_fp = [fp for fp in self.results if fp.endswith(".geojson")]
+            if json_fp:
+                json_fp = json_fp[0]  # why only one element is selected?
+            else:
+                # pylint: disable=not-an-iterable
+                json_fp = [fp for fp in self.results if fp.endswith(".json")][0]
```
rodrigoalmeida94, line 335: I guess here it would make sense to pass the whole list instead of selecting just the first element? But your question to @chrieke will probably clarify

up42/viztools.py
```diff
@@ -308,29 +322,42 @@ def map_results(
...
+            f_paths = self.results
+        elif isinstance(self.results, dict):
+            # pylint: disable=unsubscriptable-object
+            json_fp = self.results["merged_result"][0]
```
nekhtiari, line 339: Nice, so here we do not need anymore to get rid of merging results in JobCollection as @chrieke mentioned? right?

up42/viztools.py
```diff
@@ -308,29 +322,42 @@ def map_results(
...
-        if json_fp:
-            json_fp = json_fp[0]
-        else:
+        f_paths = []
+        if isinstance(self.results, list):
+            # Add features to map.
+            # Some blocks store vector results in an additional geojson file.
             # pylint: disable=not-an-iterable
-            json_fp = [fp for fp in self.results if fp.endswith(".json")][0]
+            json_fp = [fp for fp in self.results if fp.endswith(".geojson")]
```
markusuwe, line 330: Can this and the next lines be simplified into one list comprehension? E.g. by using "if os.path.splitext(fp)[0] in ["json", "geojson"] or similar.

up42/viztools.py
marooned20: Ideally, that's what I want to achieve. There has been no change in this part of the code from before. Hoping that, with some context it would bring more clarity!

markusuwe (approved):
LGTM

up42/viztools.py
marooned20: I mean, we have access to the merged results, so better to leverage that!

up42/viztools.py
marooned20: I agreed. I would wait for @chrieke and understand the context before engaging in a refactor rampage! :D

up42/viztools.py
nekhtiari: This is important not to create an extra ticket for removing that method then. Since it will then affect this one as well :D

up42/viztools.py
chrieke: E.g. the ais blocks have a data.json file with the aoi geometry, and a "ship-detection.geojson" (names differ) file with the actual results geometries. So in that case we want to prioritize the geojson file. Getting the first element is just to get rid of the list.

up42/viztools.py
chrieke: see comment above, the geojson has higher priority, so needs to be handled separatly.

up42/viztools.py
chrieke: For each job result there is just a single json (and geojson), and we only want to visualize one of them. So getting the first element just gets rid of the list. Can also do that later in the code if it makes things easier.

up42/viztools.py
chrieke: For the visualization we decided to iterate over the jobs and visualize each result independently, so the merged result is not required here. We might remove the merged result alltogether, but separate topic. I think here you need to iterate over the list of n job results instead and create n map objects.

chrieke (commented):
Added some comments for clarification of the ticket discussion.

up42/viztools.py
nekhtiari: I guess in the case of the osm block, for instance, you can have multiple .geojson file each with different geometry.

chrieke: hm okay, that makes it more complicated, since this case is also not covered in the current code. Any more blocks that have multiple json/geojson files? If too complicated maybe create a separate ticket @marooned20, might require some further discussion.

up42/viztools.py
marooned20: Thanks. I understand it now. Currently,  it handles the specific case (i.e. ship-detection). Plus, as @nekhtiari mentioned about the scenario where a block outputs a `.geojson` (e.g. OSM). I propose we have a rather generic solution for plotting/mapping functionality that handles most (if-not all/unforeseen ones) cases! 
cc: @chrieke @rodrigoalmeida94 @markusuwe

up42/viztools.py
marooned20: @chrieke  What would be the plus (and minuses) for now using `merged` result?
From current state of code and my perspective, having this file is proving to be rather beneficial. This enables us to create a single dataframe without looping and appending business, which then is used for viz.

chrieke: To clarify, merged is the merged result of all jobs, not of multiple jsons of one job.

If we merge the result and visualize only 1 map, this is at hand:
- We need to merge together also the geojson files, currently only data.json is merged together.
- Would blow up a single visualization very quickly, memory, code & usability wise (Long list of features, aois all over the globe etc.). See the ticket discussion. We decided to forego this question and do option, iterating over each job and visualizing it separatly, having multiple maps. 
- OSM block situation with multiple geojsons is new to me and not handled in the current code. Should have a separate discussion probably.

tests/test_workflow.py
```diff
@@ -708,6 +708,7 @@ def test_get_jobs_live(workflow_live):
     jobcollection = workflow_live.get_jobs()
     assert isinstance(jobcollection, list)
     assert isinstance(jobcollection.jobs[0], Job)
+    # pylint: disable=use-a-generator
     assert all(
         [j._info["workflowId"] == workflow_live.workflow_id for j in jobcollection.jobs]
     )
```
rodrigoalmeida94, line 714: You can avoid this disable by removing the list inside all. i.e.
`assert all(j._info["workflowId"] == workflow_live.workflow_id for j in jobcollection.jobs)`

marooned20: thanks! Done!

nekhtiari (approved):

chrieke (approved):
LGTM!

PR #186, opened 2021-01-22
Add ordering up5192 2/3

rodrigoalmeida94:
PR 2 will include constructors from other Classes (i.e. Catalog -> Order and Asset -> Workflow).

Constructors:
- [x] Catalog -> Order
- [x] Asset -> Workflow

Items:
* [x] Ran test & live-tests
* [x] Removed credentials

### DIFF
--- a/up42/asset.py
+++ b/up42/asset.py
@@ -47,6 +47,15 @@ class Asset(Tools):
         self._info = response_json["data"]
         return response_json["data"]
 
+    @property
+    def source(self) -> Dict:
+        """
+        Gets the source of the Asset. One of `TASKING`, `ORDER`, `BLOCK`, `UPLOAD`.
+        """
+        source = self.info["source"]
+        logger.info(f"Asset source is {source}")
+        return source
+
     def _get_download_url(self) -> str:
         url = f"{self.auth._endpoint()}/workspaces/{self.workspace_id}/assets/{self.asset_id}/downloadUrl"
         response_json = self.auth._request(request_type="GET", url=url)
--- a/up42/catalog.py
+++ b/up42/catalog.py
@@ -380,7 +380,8 @@ class Catalog(VizTools, Tools):
             scene (Series): A single item/row of the result of `Catalog.search`. For instance, search_results.loc[0]
                 for the first scene of a catalog search result.
             track_status (bool): If set to True, will only return the Order once it is `FULFILLED` or `FAILED`.
-            report_time (int): The intervall (in seconds) when to get the order status, if `track_status` is set to True.
+            report_time (int): The intervall (in seconds) when to get the order status, 
+                if `track_status` is set to True.
 
          Warning:
             When placing orders of items that are in archive or cold storage,
--- a/up42/catalog.py
+++ b/up42/catalog.py
@@ -347,7 +347,7 @@ class Catalog(VizTools, Tools):
         Args:
             geometry (Union[ Dict, Feature, FeatureCollection, List, GeoDataFrame, Point, Polygon, ]): The intended
                 output AOI of the order.
-            scene (GeoSeries): A single item/row of the result of `Catalog.search`. For instance, search_results.loc[0]
+            scene (Series): A single item/row of the result of `Catalog.search`. For instance, search_results.loc[0]
                 for the first scene of a catalog search result.
 
         Returns:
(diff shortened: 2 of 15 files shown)


up42/asset.py
```diff
@@ -47,6 +47,15 @@ def info(self) -> Dict:
...
         return response_json["data"]
 
+    @property
+    def source(self) -> Dict:
+        """
+        Gets the source of the Asset. One of `TASKING`, `ORDER`, `BLOCK`, `UPLOAD`.
+        """
+        source = self.info["source"]
+        logger.info(f"Asset source is {source}")
+        return source
```
thaisbendixen, line 57: So these are the type of assets we would have in user storage right? But we only want the asses output by blocks to be reacheable via asset id according to line 354. Why is that? Is it because everything under TASKING, ORDER and UPLOAD hasn't been fully processed and finalised, hence is not available?

thaisbendixen (approved):
LGTM!!

up42/asset.py
rodrigoalmeida94: No - the source of assets looks like it is static - that means it depends on how it was created. If an order comes from our order service it will have "ORDER" as source, if it comes from a previously run download block it will have "BLOCK" as source, regardless if it is completed or not. The issue here is that we do not support inputing all assets into workflows - assets that are not "BLOCK" source don't have a data.json and because of this we can not use them for now in a workflow. Hope this makes sense?

thaisbendixen: Ahn! Makes a lot of sense! Thanks for clarifying :)

chrieke: Good to know, thanks for question & answer.

up42/catalog.py
```diff
@@ -283,3 +285,114 @@ def download_quicklooks(
...
+        track_status: bool = False,
+        report_time: int = 120,
+    ) -> "Order":
+        """
+        Place an order from an item/row in a result of `Catalog.search`.
+
+        Args:
+            geometry (Union[ Dict, Feature, FeatureCollection, List, GeoDataFrame, Point, Polygon, ]): The intended
+                output AOI of the order.
+            scene (GeoSeries): A single item/row of the result of `Catalog.search`. For instance, search_results.loc[0]
```
chrieke, line 380: Maybe mention the required columns or assert even?

Love dataframes and having to provide the provider_id and correct provider explicitly probably leads to problems, but the dataframe row also doesn't feel right, especially since it's a single item only. Tough. Guess still best solution. Will play around a bit with this when merged before release.

up42/catalog.py
```diff
@@ -283,3 +285,114 @@ def download_quicklooks(
...
+                for the first scene of a catalog search result.
+            track_status (bool): If set to True, will only return the Order once it is `FULFILLED` or `FAILED`.
+            report_time: The intervall (in seconds) when to get the order status, if `track_status` is set tp True.
+
+         Warning:
+            When placing orders of items that are in archive or cold storage,
+            the order fulfillment can happen up to **24h after order placement**.
+            In such cases,
+            please make sure to set an appropriate `report_time`.
```
chrieke, line 389: Add "Or track the status later on via `order.track_status()`" ?

chrieke (approved):
LGTM, nice!

up42/catalog.py
rodrigoalmeida94: Not sure I fully understand what you mean - are you saying you expect users to build up their geodataframe/series from scratch and pass it on to the method? 

To create an order we need a provider and a catalog item id (plus the geometry that comes from the user). The catalog item id comes only from the catalog search (i.e. you can't get it anywhere else as far as I know).

My understanding was that we made a conscious decision to return a GeoDataFrame (by default) when we perform a catalog search. With that, following methods that require "search results" should also handle either a GeoDataFrame or some child of it. In this case since an order can only be placed for a single "search result" I would argue it's fair to say we require a row of this data frame. In the future we could also think of making some method to iterate over "search results" and place orders, but I feel that is premature right now. 

Toughts?

up42/catalog.py
rodrigoalmeida94: Well I would say if you're ordering from archive you probably won't use track_status at all no? You'd just order and later on check the status? What do you think?

up42/catalog.py
chrieke: No, I agree with that giving the dataframe row is the best option here, and it is the default result from catalog search anyway. But since you have to provide only a single dataframe row, and giving a dataframe row as an argument is a bit unusual, I was thinking if it would maybe be more explicit to just have the user provide the `provider_id` and `provider` explicitly. But that would lead to other problems (user provides the scene_id etc. because he doesn't know the difference), and I also missed that the geometry is taken from the dataframe row.

So in the end agree with this solution, just wanted to communicate some concerns, maybe you had an idea for another solution etc.

up42/catalog.py
chrieke: Yeah, I imagined it like that. Someone firing off the order, and checking 24h later via code if it is complete. Optional, but could give here a hint or code reference linking how to do that.

rodrigoalmeida94: Added

PR #202, opened 2021-03-11
s5 jupyter notebook

thaisbendixen:
Adds simple Sentinel-5p L3 (GeoTiff) workflow.

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

New visualization
Screenshot 2021-03-15 at 11 45 42


examples/data-block-examples/esa-sentinel5p-l3.ipynb
```diff
@@ -0,0 +1,347 @@
...
+      "2021-03-11 15:57:16,291 - Added tasks to workflow: [{'name': 'esa-s5p-l3-gtiff:1', 'parentName': None, 'blockId': '70919851-c563-4e5e-9393-029cf10377c7'}]\n",
+      "2021-03-11 15:57:16,291 - Getting small example aoi in location 'Berlin'.\n"
+     ]
+    }
+   ],
+   "source": [
+    "input_tasks = [\"esa-s5p-l3-gtiff\"]\n",
+    "workflow.add_workflow_tasks(input_tasks)\n",
+    "# Define the aoi \n",
+    "aoi = up42.get_example_aoi(as_dataframe=False)\n",
```
nekhtiari, line 72: I guess you don't need this since you specify your own aoi in next cell, right?

nekhtiari (approved):

examples/data-block-examples/esa-sentinel5p-l3.ipynb
thaisbendixen: Good catch!

chrieke (commented):
`workflow.construct_parameters` should already add the additional default parameters of the block. If it was intended to show setting them manually also good.
For the visualization you could select the `cmap` plotting parameter, e.g. red or blue etc. Doesn't make it look much better but the black/white looks a bit like an error. Otherwise LGTM

markusuwe (commented):
The visualisations are not that great, they are only black and white. Can we have something more pretty?
As in the other example, if the output of a task is getting extensive it is better do delete i.

marooned20 (approved):

PR #203, opened 2021-03-11
s1 slc sobloo example

thaisbendixen:
Adds simple Sentinel-1 SLC (Fullscene) workflow. Unfortunatly it's not possible to plot quicklook without path because of the SAFE file structure.

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


chrieke (commented):
Nice with the quicklook plot. Could you add a ticket to the SDK epic please that for SAFE quicklook doesn't work out of the box, think we can fix this.

Also would maybe delete the longt output of the download-cell, is not super important. Otherwise LGTM!

markusuwe (approved):
LGTM
Agree with @chrieke 's comments.

PR #215, opened 2021-03-29
Add pleiades-download reflectance data example

Onyothi:
Add pleiades reflectance download data example using catalog search_UP-5271
https://up42.atlassian.net/browse/UP-5271
Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


thaisbendixen (commented):
Nice job with doing the catalog search! Some suggestions:
- Remove the commented display(search_results.head()) line of you are not using
- Remove the overly long track cells from workflow.run_job and job.download_results()
- Update the comment in cell 28 "Select least cloud scene for further workflow." -> "Select least cloudy scene for further workflow."

rodrigoalmeida94 (commented):
Some notes:
- Seems like all the other notebooks have the example termination. I would suggest you change the file name to pleiades-download-example.ipynb.
- Can we also add a visualisation of the image at the end? I guess loading the XML file in rasterio for the MS band should do the trick.
- Would also be really great to add a section where you describe how you can now reuse this data making use of the asset id mechanism. This is also mentioned in the [docs](https://docs.up42.com/going-further/download-blocks-tutorial-spot-pleiades.html#how-to-re-use-the-downloaded-image), but would be great to have a programatic "live" example.

Onyothi:
> * docs

rodrigoalmeida94 (approved):
Nice job! Just one final improvement:
You don't need to do this the step that reads `# Read the RGB bands in memory and append them to a list.`. There you are creating a list from a list. 
You can pass the list of bands to dstack directly:
`rgb = np.dstack([red, green, blue])`

markusuwe (approved):
Nice!

PR #224, opened 2021-04-20
Deprecate tools inheritance

chrieke:
Fully deprecates the Tools module inheritance  in the other classes (after deprecation process). The tools functions will only be available from the `up42` import object.

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/up42/auth.py
+++ b/up42/auth.py
@@ -219,7 +219,7 @@ class Auth:
         data: Union[dict, list] = {},
         querystring: dict = {},
         return_text: bool = True,
-    ) -> Union[str, dict, requests.Response]:
+    ):  # Union[str, dict, requests.Response]:
         """
         Handles retrying the request and automatically retries and gets a new token if
         the old is invalid.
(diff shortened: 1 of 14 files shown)


up42/auth.py
```diff
@@ -219,7 +219,7 @@ def _request(
         data: Union[dict, list] = {},
         querystring: dict = {},
         return_text: bool = True,
-    ) -> Union[str, dict, requests.Response]:
+    ):  # Union[str, dict, requests.Response]:
```
chrieke, line 222: Not exactly sure why only the removed class inheritance triggered lots of type checks issues of the response, but had to deactivate for now. The current potential API return types are not ideal (json, text, quicklook data chunkz), this will at some point be adressed at API level.

markusuwe: Was looking at the problem before and could not make sense of it. Don't think it's critical, but it would surely be good if that would be resolved some time in the future; as this repo is open source people will wonder. Maybe create a tech ticket?

markusuwe (approved):
LGTM

PR #229, opened 2021-04-21
Improve retry

chrieke:
- Small modification of retry strategy
- Small docstring & code improvements in auth module

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/up42/auth.py
+++ b/up42/auth.py
@@ -25,9 +25,8 @@ logger = get_logger(__name__)
 
 class retry_if_429_error(retry_if_exception):
     """
-    Retry strategy that retries if the exception is an ``HTTPError`` with
-    a 429 status code (too many requests).
-
+    Altered tenacity retry strategy that retries if the exception is an ``HTTPError``
+    with a 429 status code (too many requests).
     Adapted from https://github.com/alexwlchan/handling-http-429-with-tenacity
     """
 
@@ -66,6 +65,7 @@ class Auth:
         self.project_id = project_id
         self.project_api_key = project_api_key
         self.workspace_id: Optional[str] = None
+        self.token: Optional[str] = None
 
         try:
             self.env: str = kwargs["env"]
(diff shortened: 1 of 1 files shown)


up42/auth.py
```diff
@@ -247,8 +249,8 @@ def _request(
         try:
             if self.retry:
                 retryer = Retrying(
-                    stop=stop_after_attempt(1),  # TODO: Find optimal retry solution
-                    wait=wait_fixed(0),
+                    stop=(stop_after_attempt(2) | stop_after_delay(3)),
```
markusuwe, line 252: Isn't this redundant? If 2 attempts are allowed and we wait 0.5 seconds for each of them, then the retry will stop after 1 s if If understand the different parameters correctly.

markusuwe (approved):
LGTM

up42/auth.py
```diff
@@ -42,9 +43,9 @@ def is_http_429_error(exception):
 class Auth:
     def __init__(
         self,
-        cfg_file: Union[str, Path, None] = None,
```
rodrigoalmeida94, line 45: Is this change in order absolutely necessary? I think this could break users workflows in case they make use of order of arguments instead of keyword arguments.

up42/auth.py
rodrigoalmeida94: I also feel the same, I guess a single stop method should be sufficient in this.

up42/auth.py
chrieke: Reverted. Think it would be good to have the config file as the second option, but true, maybe better suited when going to 1.0.0.

up42/auth.py
chrieke: Thought about it as a fallback, we do 2 retries, but quit if they take longer than 3 seconds, e.g. if there is a server issue? But maybe the 2 retries should just play out, will remove the second method.

PR #235, opened 2021-06-18
Add terra cove lake monitoring example jupyter notebook

thaisbendixen:
This PR addresses the following ticket https://up42.atlassian.net/browse/UP-5532?atlOrigin=eyJpIjoiMmRiYmU4MWE0NzE2NGFlOTg0MzUyYWQ4NDRkMzllMWEiLCJwIjoiaiJ9.

This example was prepared by Terra cover and highlights how their block can be used to monitor shrinking lakes. The workflow takes 1 hours to run, which is mentioned in the introduction of the example. This workflow example highlights the blocks usefulness without having to run the workflow.

The workflow can't be shortened, as It checks the lake sizes for several days and two lakes (the time and number of lakes already has been reduced significantly).

The notebook is now adapted to a rolling archive and looks at teh previous

For release:
* [ ] Add links to blocks in marketplace

For release:
* [ ] Bumped version
* [ ] Added changelog


chrieke (commented):
Cool example! Even though a notebook coming from a partner could be good if we improve readability a bit. 

- The long `for wbctr in range(len(water_body_ids)):` loop in the visualization section is copied 3 times.
- The "# dividing the scenes into batches to submit multiple jobs for faster execution" sections could be replaced with `construct_parameters_parallel` (created a ticket to better surface parallel job examples).

To show the notebook on the docs need to add the chapter to `/mkdocs.yml` in the `nav` section, and in `/docs/examples.md` (maybe some "Partner examples" section etc?).

marooned20 (approved):
Looks good to me!

Minor suggestion: in the visualize results function there is `src.close()` after reading each bands can be removed as files are opened with context manager anyway

PR #237, opened 2021-06-29
Limit memory usage for large file downloads

j-tr:
Currently, the SDK downloads the entire file into memory before storing it on disk. This causes out-of-memory issues for downloads larger than the available RAM size (this is an actual problem for our latest order).
This can be simply mitigated by enabling stream download.


chrieke:
@j-tr Thanks! Putting out a new version later today.

PR #244, opened 2021-08-05
Accept datetime up4677

chrieke:
Add support for datetime objects and all iso-format compatible time strings to construct_parameters (start_date, end_date).

* [] After initial reviews, add to catalog.

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/workflow.py
```diff
@@ -378,9 +382,26 @@ def construct_parameters(
                 input_parameters[data_block_name]["ids"] = scene_ids
                 input_parameters[data_block_name]["limit"] = len(scene_ids)
                 input_parameters[data_block_name].pop("time")
-            elif start_date is not None and end_date is not None:
-                time = f"{start_date}T00:00:00Z/{end_date}T23:59:59Z"
-                input_parameters[data_block_name]["time"] = time
+            elif start_date is not None or end_date is not None:
+                try:
+                    start_date_iso: datetime = datetime.fromisoformat(start_date)  # type: ignore
+                    end_date_iso: datetime = datetime.fromisoformat(end_date)  # type: ignore
```
thaisbendixen, line 388: What if it's already datetime object? Will this always fail?

up42/workflow.py
```diff
@@ -378,9 +382,26 @@ def construct_parameters(
...
+                    if len(end_date) == 10:  # type: ignore
+                        # End date provided in the form "2014-01-01" without clock time, is set
+                        # to the previous day as upper boundary of archive.
+                        end_date_iso = datetime.combine(
+                            end_date_iso.date(), datetime_time(23, 59, 59)
+                        )
```
thaisbendixen, line 394: We only change the end_date time because it's not set correctly by the datetime.fromisoformat method?

thaisbendixen (approved):
LGTM! Just a few questions

up42/workflow.py
chrieke: Yes. But should maybe change this logic and instead ask for the different types to be more explicit in the code.

up42/workflow.py
chrieke: The default "todatetime" parsing from a "YYYY-MM-DD" string sets it to the beginning of the day (00:00:00), but the image query requires the end of day. Thanks for the note here, even I got confused now. Overhauled this whole section and clarified it, still not the nicest code but better than before.

thaisbendixen: Looks great now!! Thanks

thaisbendixen (approved):
LGTM!

PR #249, opened 2021-08-16
Catalog pagination up4697

chrieke:
Adds handling of catalog search results pagination token to give the user the full results items when asking for more than the default 500 limit in the search parameters.

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/catalog.py
```diff
@@ -196,12 +196,31 @@ def search(
...
-        dst_crs = "EPSG:4326"
-        df = GeoDataFrame.from_features(response_json, crs=dst_crs)
+        features = response_json["features"]
+
+        # Catalog search gives results with >500 items as 50 page pagination tokens.
+        # Pagination ends when both links are the same.
+        while len(features) < max_limit:
+            url_page = response_json["links"][1]["href"]
+            if url_page == response_json["links"][0]["href"]:
+                break
```
thaisbendixen, line 214: If this condition is not true in the first loop, it will never be true in the other loops. url_page und response_json never get renewed during the loop.

thaisbendixen (changes requested):
Really complicated scenario! Overall looks really great! Just found one bug

up42/catalog.py
chrieke: Great catch!!!!

chrieke: Added testcase for pagination exhaustion (catalog_pagination_mock only provides one time 500 results, then only one page with 50 results. Only 550 results are returned even though limit was 614).

Loop should also be more readable now with new variable names.

thaisbendixen (approved):
NOICE!

PR #250, opened 2021-08-16
remove geometry filtering for catalog search results

chrieke:
SDK used an additional geometry filtering for catalog search results, to filter results out that were not actually touching the search geomety. Instead will be ticketed with backend and removed, since it is also interfering with the catalog pagination testing.

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


thaisbendixen (approved):
LGTM!

PR #252, opened 2021-09-08
Compatible blocks empty workflow up5561

chrieke:
Fixes error with `.get_compatible_blocks()` if the workflow has no tasks defined. In this case now gives logger message and returns all data blocks, since these are all compatible for an empty workflow.

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


thaisbendixen (approved):
That was quick!!

PR #255, opened 2021-09-13
Remove squash geometry - up4761

chrieke:

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/utils.py
```diff
@@ -308,92 +308,47 @@ def any_vector_to_fc(
...
 
-    # TODO: Handle multipolygons
-
-    # With the now uniform feature collection, decide to return a feature or list of bounds (bbox).
+    geometry_error = "The provided geometry {}, UP42 only accepts single geometries."
     if len(fc["features"]) == 1:
         f = fc["features"][0]
-        if geometry_operation == "bbox":
-            try:
-                query_geometry = list(f["bbox"])
```
jmigueldelgado, line 361: I don't love this `f` dictionary... I tend to read it as an f-string at first glance and only later I realise it is actually a dictionary.

jmigueldelgado (dismissed):
LGTM. I was a bit surprised that you were removing functionalities from the sdk (support for multi-feature AOIs) but it actually makes sense. 👍

up42/utils.py
chrieke: good point, changed to feature

jmigueldelgado (approved):
👍

PR #259, opened 2021-11-09
Speed up create workflow existing up5736

chrieke:
Speed up project.create_workflow(..., use_existing=True) by using the json return of all workflows in the project to compare with the workflow that is supposed to be created. Previously the workflow object of each workflow was used for the comparison which takes much longer due to each additional workflow.info call.

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


thaisbendixen (approved):
LGTM! Still don't understand 100% how this change can make a significant difference! Let's talk about It in our dev duo

PR #283, opened 2022-01-27
Add collection support up5754

chrieke:
- Adds support for UP42 data collections
- Breaking change: Adjust catalog.construct_parameters to use collections instead of sensors (already removed from API).

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/up42/catalog.py
+++ b/up42/catalog.py
@@ -53,8 +53,8 @@ class Catalog(VizTools):
         return json_response["data"]
 
     # pylint: disable=dangerous-default-value
+    @staticmethod
     def construct_parameters(
-        self,
         geometry: Union[
             dict,
             Feature,
--- a/up42/catalog.py
+++ b/up42/catalog.py
@@ -3,7 +3,7 @@ Catalog search functionality
 """
 
 from pathlib import Path
-from typing import Union, List, Tuple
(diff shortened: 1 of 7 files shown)


up42/catalog.py
```diff
@@ -121,8 +81,8 @@ def construct_parameters(
                 list, GeoDataFrame, Point, Polygon.
             start_date: Query period starting day, format "2020-01-01".
             end_date: Query period ending day, format "2020-01-01".
-            sensors: The satellite sensors to search for, one or multiple of
-                ["pleiades", "spot", "sentinel1", "sentinel2", "sentinel3", "sentinel5p"]
+            collection: The satellite sensor collection to search for, e.g. "PHR",
```
chrieke, line 84: @mykolakozyr @iliascholl-up42 @vzup42 @dc-up42 Seeing the "Catalog Console: Search by Hosts"  story on productboard, does it make sense to already implement it in the SDK as this, a single collection name string instead of a list of collections? Or will the API still accept a list of collections after? Would prefer to not change the parameter type if this changes in a few weeks.

chrieke: After discussion with Mykola, we will keep it as a list.

PR #292, opened 2022-02-11
Add get job credits methods and tests

thaisbendixen:
The support team requested that the SDK outputs credit usage for each job. 

Two questions:
- What else would I need to update in the documenattion. The method doc is automatically uploaded to the docs. 
- What happens if the job status is "Running". This could fail right? The download_results method doesn't account for this either, so I'm wondering what the behaviour is here.

Items:
* [x] Ran test & live-tests

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/up42/job.py
+++ b/up42/job.py
@@ -349,7 +349,7 @@ class Job(VizTools):
 
     def get_job_credits(self) -> dict:
         """
-        Convenience function to get the credit consumption of the job.
+        Convenience function that gets number of credits consumed in job.
 
         Returns:
             The consumed credits for the job.
--- a/up42/job.py
+++ b/up42/job.py
@@ -332,7 +332,7 @@ class Job(VizTools):
         in a dictionary of strings.
 
         Returns:
-            The data.json of alle single job tasks.
+            The data.json of all single job tasks.
         """
         jobtasks: List[dict] = self.get_jobtasks(return_json=True)  # type: ignore
         jobtasks_ids = [task["id"] for task in jobtasks]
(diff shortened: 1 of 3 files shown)


jmigueldelgado (approved):
LGTM!

up42/job.py
```diff
@@ -346,3 +346,17 @@ def get_jobtasks_results_json(self) -> dict:
 
             jobtasks_results_json[jobtask_id] = response_json
         return jobtasks_results_json
+
+    def get_job_credits(self) -> dict:
+        """
+        Convenience function that gets number of credits consumed in job.
```
chrieke, line 352:
```suggestion
        Gets the credit costs of the job.
```

up42/job.py
```diff
@@ -346,3 +346,17 @@ def get_jobtasks_results_json(self) -> dict:
 
             jobtasks_results_json[jobtask_id] = response_json
         return jobtasks_results_json
+
+    def get_job_credits(self) -> dict:
```
chrieke, line 350: Should we lose the "job" in the function name since the object is already job? `job.get_credit_costs()` or `job.get_credits()` etc.?

For "sound" reference, we probably will add some endpoints like `up42.get_credit_history()` as well.

chrieke (commented):
LGTM! Smooth.

chrieke:
> What else would I need to update in the documenattion. The method doc is automatically uploaded to the docs.
Nothing, only if you want to add it to the examples, but think we can do that when all credits functionality is added.

> What happens if the job status is "Running". This could fail right? The download_results method doesn't account for this either, so I'm wondering what the behaviour is here.
Tested it, suprisingly while the job was still running still gave me the correct result. In any case, this should be handled via the API / API error.

PR #297, opened 2022-03-03
added asset_ids in the function explanation

teoEnache:


teoEnache:
@chrieke this only solves the problem short-term, it might be necessary to intervene in the variable names for assets

chrieke:
@teoEnache Thanks for the PR! The variable is supposed to be type `List[up42.asset.Asset]`, so e.g. what is returned by `storage.get_orders()`. With a list of asset-ids it currently would not work.
As far as I can tell see this was implemented this way to be able to validate that the provided asset ids belong to assets that were created via a block, see [here](https://github.com/up42/up42-py/blob/287958a0d121764dd2b1add83de0c371aefc33a8/up42/workflow.py#L379). But agree, this could potentially just be the ids. But we need to have a better understanding how this is used in the future, crated a ticket [INT-42](https://up42.atlassian.net/browse/INT-42?atlOrigin=eyJpIjoiNTdiMTIyMjUwYjZmNDEwZTllZmZiYmViMGVlZTg1Y2UiLCJwIjoiaiJ9).

chrieke:
Closing for now, we will address this via the Jira ticket.

PR #306, opened 2022-04-07
docs fixes

chrieke:
Addresses some of the recent feedback, small docs fixes and improvements.

- Fixes structure tabs
- Clarifies Python version in docs
- Integrates test.sh file in makefile to have a single entrypoint.
- Updates docs requirements & removes subdepencies that are installed by other packages

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


andher1802 (commented):
On the requirements-docs.txt last time I required to set jninja to 3.0.0 because they released a new version on March that breaks medics. Only to mention, if we do not require anymore ninja it should be OK.

andher1802 (approved):

andher1802:
LGTM, only one small comment on the requirements to make sure that not breaks anything. Nice to remove test.sh, I think less is more. Good job.

chrieke:
> On the requirements-docs.txt last time I required to set jninja to 3.0.0 because they released a new version on March that breaks medics. Only to mention, if we do not require anymore ninja it should be OK.

Yeah was also suprised to see that it works, I guess mkdocs-material fixed something. Tested with a new env and mkdocs install from scratch (installs jinia). Now pinned newest mkdocs-material.

PR #308, opened 2022-04-08
bumping to V0.20.2

andher1802:

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ X] Added changelog


chrieke (commented):
Since these are all non functional changes not affecting the user I think we should avoid confusion with too much detail. Something like "Non functional changes to enable conda release" etc?

chrieke (approved):

PR #311, opened 2022-04-19
Adding get_credits_balance and get_credits_history

andher1802:
Adding the get_credits_balance and get_credits_history on the tools Class.

Items:
* [ x] Ran test & live-tests
* [ x] Implemented (new) unit tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -308,24 +308,33 @@ def credits_history_mock(auth_mock, requests_mock):
         url=url_get_credits_history,
         json=response_json,
     )
-    # Pagination (from 1 to 5)
+    return Tools(auth=auth_mock)
+
+@pytest.fixture()
+def credits_history_pagination_mock(auth_mock, requests_mock):
+    with open(
+        Path(__file__).resolve().parent / "mock_data/credits_history.json"
+    ) as json_file:
+        credits_history_response_json = json.load(json_file)
+
+    response_json = copy.deepcopy(credits_history_response_json)
     temp_content = copy.deepcopy(response_json["data"]["content"])
     pagination_response_json = copy.deepcopy(response_json)
     pagination_response_json["data"]["last"] = False
-    pagination_response_json["data"]["content"] = temp_content[:5]
+    pagination_response_json["data"]["content"] = temp_content * 200
     url_get_credits_history_size = (
-        f"{auth_mock._endpoint()}/accounts/me/credits/history?size=5&page=0"
+        f"{auth_mock._endpoint()}/accounts/me/credits/history?page=0"
     )
     requests_mock.get(
         url=url_get_credits_history_size,
         json=pagination_response_json,
     )
-    # Pagination (from 5 to 10)
+    # Page 2
     pagination_response_json_last = copy.deepcopy(response_json)
     pagination_response_json_last["data"]["last"] = True
-    pagination_response_json_last["data"]["content"] = temp_content[5:]
+    pagination_response_json_last["data"]["content"] = temp_content * 50
     url_get_credits_history_size_last = (
-        f"{auth_mock._endpoint()}/accounts/me/credits/history?size=5&page=1"
+        f"{auth_mock._endpoint()}/accounts/me/credits/history?page=1"
     )
     requests_mock.get(
         url=url_get_credits_history_size_last,
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -302,11 +302,34 @@ def credits_history_mock(auth_mock, requests_mock):
         Path(__file__).resolve().parent / "mock_data/credits_history.json"
     ) as json_file:
         credits_history_response_json = json.load(json_file)
-    pagination_response_json = copy.deepcopy(credits_history_response_json)
+    response_json = copy.deepcopy(credits_history_response_json)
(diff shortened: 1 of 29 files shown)


andher1802:
@chrieke this is a preliminary PR for keep you updated about progress and possible feedback for further reference. I need to improve the testing and also add local tests to the get_credits_history function.

tests/test_tools.py
```diff
@@ -104,6 +105,55 @@ def test_get_block_details_live(tools_live):
...
+        start_date="2022-03-01", end_date="2022-03-02"
+    )
+    assert isinstance(balance_history, dict)
+    assert "content" in balance_history
+    assert len(balance_history["content"]) == 32
+    assert isinstance(balance_history["content"], list)
+
+
+@pytest.mark.live
+def test_get_credits_history_no_bounds_live(tools_live):
```
chrieke, line 142: What this is really checking is that the function code can handle a not defined start (or enddate), not so much the content of the response. Therefore could also just be a mocktest. Ideally could be using @pytest.parameterize .

tests/test_tools.py
```diff
@@ -104,6 +105,55 @@ def test_get_block_details_live(tools_live):
...
+
+
+@pytest.mark.live
+def test_get_credits_history_live(tools_live):
+    balance_history = tools_live.get_credits_history(
+        start_date="2022-03-01", end_date="2022-03-02"
+    )
+    assert isinstance(balance_history, dict)
+    assert "content" in balance_history
+    assert len(balance_history["content"]) == 32
```
chrieke, line 137: Could be that in the future this fails if test runs add additional items to the credit history of the sdk test account, maybe do `>= 32` Allthough weird that it is only 32 overall hm.

up42/tools.py
```diff
@@ -224,6 +223,88 @@ def get_block_details(self, block_id: str, as_dataframe: bool = False) -> dict:
...
+        account.
+        Args:
+            start_date: The start date for the credit consumption search e.g.
+            2021-12-01. Default start_date None uses 2000-01-01.
+            end_date: The end date for the credit consumption search e.g.
+            2021-12-31. Default end_date None uses current date.
+
+        Returns:
+            A dict with the information of the credit consumption records for
+            all the users linked by the account_id. The result match the endpoint
```
chrieke, line 259: "The result match the endpoint response with all records on the content key."
Not sure what this means, possible to clarify or remove?

up42/tools.py
```diff
@@ -224,6 +223,88 @@ def get_block_details(self, block_id: str, as_dataframe: bool = False) -> dict:
...
+            end_date = tomorrow_datetime.strftime("%Y-%m-%d")
+
+        [start_formatted_date, end_formatted_date] = format_time_period(
+            start_date=start_date, end_date=end_date
+        ).split("/")
+        search_parameters = dict(
+            {
+                "from": start_formatted_date,
+                "to": end_formatted_date,
+                "size": 2000,
```
chrieke, line 286:
```suggestion
                "size": 2000,  # 2000 is the maximum page size for this call
```

up42/tools.py
```diff
@@ -224,6 +223,88 @@ def get_block_details(self, block_id: str, as_dataframe: bool = False) -> dict:
...
+        )
+        endpoint_url = f"{self.auth._endpoint()}/accounts/me/credits/history"
+        response_json: dict = self.auth._request(
+            request_type="GET", url=endpoint_url, querystring=search_parameters
+        )
+        isLastPage = response_json["data"]["last"]
+        credit_history = response_json["data"]["content"]
+        result = dict(response_json["data"])
+        del result["content"]
+        while not isLastPage:
```
chrieke, line 298: We should probably add a test for the pagination, a bit similar to [here](https://github.com/up42/up42-py/blob/88a34b05d4606811943dd41e10078f29da551d3c/tests/test_catalog.py#L290), also see the pagination_mock fixture.

chrieke (commented):
Nice! A few small comments. Probably good to add a mock test for the pagination, as due to the low order volume on the sdk test account this can not be tested in live scenario.

tests/test_tools.py
andher1802: I fixed the inputs for being just one day. Therefore, I think we should not expect changes as there could not be more transactions on this time range. But, maybe if there is some adjust on the database and they include new credits consumed on this date, then we could have an error.

chrieke: ah ja, forgot about the fixed dates.

tests/fixtures.py
```diff
@@ -234,6 +234,11 @@
     },
     "error": None,
 }
+
+JSON_BALANCE = {"data": {"balance": 10693}}
+
+JSON_CREDITS_HISTORY = JSON_ORDERS
```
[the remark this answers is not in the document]
andher1802: Done!

tests/test_tools.py
andher1802: Done!

tests/test_tools.py
andher1802: Checked and Done!

up42/tools.py
andher1802: Done! Not the same approach than the catalog pagination test, but similar idea.

PR #365, opened 2022-10-20
add parameter suggestions to construct_order_parameters

chrieke:
Uses data_product_schema to validate the parameters and logs suggestions for missing parameters used in catalog.construct_order_parameters

Same functionality in tasking module:
CleanShot 2022-10-23 at 23 34 47@2x

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/catalog.py
```diff
@@ -358,54 +359,58 @@ def search(
...
+            list,
+            GeoDataFrame,
+            Polygon,
         ] = None,
     ):
         """
-        Helps constructing the parameters dictionary required for the catalog order.
+        Helps constructing the parameters dictionary required for the catalog order. Some datasets have
+        additional parameters that are added to the output dictionary with value None. The potential values to
+        select from are given in the logs, for more detail on the parameter use `catalog.get_data_product_schema()`.
```
andher1802, line 378: catalog.get_product_schema is an endpoint, should we reference the endpoint better?. this method does not have too much information.

up42/catalog.py
```diff
@@ -358,54 +359,58 @@ def search(
...
             aoi = any_vector_to_fc(vector=aoi)
             aoi = fc_to_query_geometry(fc=aoi, geometry_operation="intersects")
-            order_parameters["params"]["aoi"] = aoi  # type:ignore
+            params["aoi"] = aoi  # type:ignore
+
+        logger.info(
+            "See `catalog.get_data_product_schema(data_product_id)` for more detail on the parameter options."
+        )
+        schema = self.get_data_product_schema(data_product_id)
+        order_parameters = autocomplete_order_parameters(
```
andher1802, line 411: I like the idea of making easier to the user for creating the order parameters by autocompleting them with a schema.

andher1802 (approved):
in general it LGTM!.

One question, I tried to complete an example calling for the autocomplete_order_parameters. In the complete case (with AOI) it resulted in no log suggestions. In the incomplete case (missing the AOI) it resulted in suggesting me the AOI polygon (as expected). However, I was expecting to suggesting me values based on the schema from the endpoint (in this case: 
'definitions': {'Polygon': {'title': 'Polygon',
   'description': 'Polygon Model',
   'type': 'object',
   'properties': {'coordinates': {'type': 'array',
     'items': {'type': 'array',
      'minItems': 4,
      'items': {'type': 'array', 'minItems': 2, 'items': {'type': 'number'}}}},
    'type': {'title': 'Type', 'const': 'Polygon', 'type': 'string'}},
   'required': ['coordinates']}},
 'additionalProperties': False,
 'title': 'OneAtlasParams'}).

Should I expect to get suggestions for title, or it only suggests in the case there is additionalProperties?. Maybe something to add in the docs for dev reference. 

Nice work! 
 
CompleteCase
IncompleteCase

up42/catalog.py
chrieke: Added all three variants to the unit test.

up42/catalog.py
chrieke: How do you mean? Explain what it does here? Hm not sure, it should just be a small note that this function might be relevant, has its own docstring.

chrieke:
> in general it LGTM!.
> 
> One question

Yeah, I wasn't sure about including the whole Geometry definition, since it obfuscates the logs. Also I'm not sure if a regular user really would know which format to provide from the schema. Was thinking about rather adding a manual example to the logs ala "Provide a Polygon   {'type': 'Polygon', 'coordinates': (((lat, lon), ...)}"   etc? WDYT?

PR #384, opened 2022-11-01
Handle required non default workflow.construct_parameters

chrieke:
Changes behaviour of `workflow.construct_parameters` to be more user friendly.

- In workflow.construct_parameters, adds any parameters with `None` default, previously only non-None parameters were added.
- If a parameter is required but has no default value, now adds that value as `None` to inform the user about this parameter. Previously parameters without defaults were not added.

- Refactor `_get_default_parameters`
- Adds test cases for the different parameter conditions
- Fixes fixture values which were "None" strings instead of actual strings

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


andher1802 (approved):
I don't see anything out of order. LGTM!
mostly changes on parameters for tests "None" -> None
And changes on workflow.py to enforce adding default as placeholder None if it does not exists instead of ignoring the parameter.

PR #396, opened 2023-01-11
update sonarcloud context

seedlit:
Updated sonarcloud context.
Previous context and token will be deleted.


chrieke (approved):
thank you!

PR #411, opened 2023-02-08
Adding pagination storage

andher1802:

Items:
* [x ] Ran test & live-tests
* [] Implemented (new) unit tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/up42/storage.py
+++ b/up42/storage.py
@@ -78,6 +78,26 @@ class Storage:
             results_list += response_json["content"]
         return results_list[:limit]
 
+    def _query_paginated_stac_search(
+        self, 
+        url: str, 
+        stac_search_parameters: dict,
+        token: Optional[str] = None
+        ) -> list:
+
+        response_features = []
+        while True:
+            stac_results = self.auth._request(
+                request_type="POST", url=url, data=stac_search_parameters
+            )
+            response_features.extend(stac_results["features"])
+            link_elements = [link["body"]["token"] for link in stac_results["links"] if link["rel"] == "next"]
+            if (len(link_elements) > 0):
+                stac_search_parameters["token"] = link_elements[0]
+                continue
+            else:
+                return response_features
+
     def _search_stac(
         self,
         acquired_after: Optional[Union[str, datetime]] = None,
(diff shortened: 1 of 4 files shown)


up42/storage.py
```diff
@@ -78,6 +78,34 @@ def _query_paginated(
...
+        response_features = []
+        while True:
+            stac_results = self.auth._request(
+                request_type="POST", url=url, data=stac_search_parameters
+            )
+            response_features.extend(stac_results["features"])
+            link_elements = [link["body"]["token"] for link in stac_results["links"] if link["rel"] == "next"]
+            if (len(link_elements) > 0):
+                stac_search_parameters["token"] = link_elements[0]
+                continue
```
chrieke, line 105:
```suggestion
                break
```

up42/storage.py
```diff
@@ -78,6 +78,34 @@ def _query_paginated(
...
+            else:
+                return response_features
```
chrieke, line 107: return should be outside of the loop, how about:

        while True:
            stac_results = self.auth._request(
                request_type="POST", url=url, data=stac_search_parameters
            )
            response_features.extend(stac_results["features"])
            token_list = [link["body"]["token"] for link in stac_results["links"] if link["rel"] == "next"]
            if not token_list:
                break
            stac_search_parameters["token"] = link_elements[0]

     return response_features

up42/storage.py
```diff
@@ -78,6 +78,34 @@ def _query_paginated(
...
+
+        Args:
+            url (str): The base url for paginated endpoint.
+            stac_search_parameters (dict): the parameters required for stac search
+
+        Returns:
+            List of storage STAC results features.
+        """
+        response_features = []
+        while True:
```
chrieke, line 97: `while` loop obviously dangerous. Even though the logic seems save, we could in theory add another precaution:

Since we have `max_items` and `limit` in the search_parameters, we know that there can be a maximum of `limit/max_items` requests.

chrieke (approved):
LGTM, nice!

Issue #357, opened 2022-10-12
Upgrade mkdocstrings to 0.20.x


chrieke:
mkdocstrings 0.19.0 does not support showing inherited_members yet, thus we downgraded to 0.18.1 (especially relevant for showing catalog base class functions) https://github.com/up42/up42-py/pull/356

This is a reminder issue: 
- Upgrade mkdocstrings to 0.20.x
- In up42 reference module use as `up42` instead `up42.__init__`
- In code reference md files, merge `selection` and `rendering` into `options`
- Check correct members inherited
- Check class properties and functions sepaerated in code reference

PR #418, opened 2023-02-14
Updating privacy policy to point to the UP42 link.

andher1802:
Pointing the privacy policy to the main UP42 privacy policy page https://up42.com/legal/privacy
Adding the mkdocstrings-python requirement to avoid the python handler error.

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials
* [X] Updated [documentation](sdk.up42.com)

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,5 +1,6 @@
 mkdocs-material==8.5.11
 mkdocstrings==0.19.1
+mkdocstrings-python
 mkdocs-exclude==1.0.2
 mkdocs-jupyter==0.22.0
 mkdocs-autolinks-plugin==0.6.0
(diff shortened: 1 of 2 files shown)


requirements-docs.txt
```diff
@@ -1,5 +1,6 @@
 mkdocs-material==8.5.11
 mkdocstrings==0.19.1
+mkdocstrings-python
```
chrieke, line 3: Can you check your mkdocstrings & mkdocs-material version? This should only be required with newer mkdocstrings versions, not 0.19.1 yet. See issue https://github.com/up42/up42-py/issues/357 why this is still pinned to older version.

chrieke (approved):
LGTM! (see dependency issue comment)

requirements-docs.txt
andher1802: I think that my version is 0.19.1, but checking the docs I think the python handler is required from version 0.18.1.

Screen Shot 2023-02-17 at 10 48 39

https://mkdocstrings.github.io/handlers/overview/

On my env it did not work (but it worked before), therefore I am not sure if they have changed something. 
Can you check if make serve work for you?

PR #452, opened 2023-08-08
DOC-652: Remove OSM documentation

dustgalactic:
The **OpenStreetMap** data block is getting sunset. This PR is removing the notebooks that use this block.


andher1802 (approved):
lgtm

PR #463, opened 2023-09-04
[AT-3079] Add functionality to download STAC assets

seedlit:
Jira Ticket [here](https://up42.atlassian.net/browse/AT-3079).
Scope: Adds functionality to download STAC asset on up42's Asset object.

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/tests/fixtures/fixtures_asset.py
+++ b/tests/fixtures/fixtures_asset.py
@@ -1,8 +1,7 @@
 import datetime
 import os
-
-import pytest
 from pathlib import Path
+import pytest
 from pystac import Item, ItemCollection
 from pystac.collection import Extent, SpatialExtent, TemporalExtent
 from pystac_client import CollectionClient
--- a/tests/fixtures/fixtures_asset.py
+++ b/tests/fixtures/fixtures_asset.py
@@ -1,6 +1,5 @@
 import datetime
 import os
-from pathlib import Path
 import pytest
 from pystac import Item, ItemCollection
 from pystac.collection import Extent, SpatialExtent, TemporalExtent
@@ -12,7 +11,6 @@ from .fixtures_globals import (
     ASSET_ID2,
     DOWNLOAD_URL,
     DOWNLOAD_URL2,
-    STAC_ASSET_DOWNLOAD_URL,
     JSON_ASSET,
     JSON_STORAGE_STAC,
     STAC_COLLECTION_ID,
@@ -157,7 +155,7 @@ def asset_mock2(auth_mock, requests_mock):
     )
     asset = Asset(auth=auth_mock, asset_id=ASSET_ID2)
     return asset
-    
+
 
 @pytest.fixture()
 def asset_live(auth_live):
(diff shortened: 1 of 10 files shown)


tests/fixtures/fixtures_asset.py
```diff
@@ -155,7 +157,7 @@ def asset_mock2(auth_mock, requests_mock):
     )
     asset = Asset(auth=auth_mock, asset_id=ASSET_ID2)
     return asset
-
+
```
andher1802: Always happens to me, unwanted tabs.

up42/stac_client.py
```diff
@@ -3,7 +3,7 @@
 from up42.auth import Auth
 
 
-class pystac_auth_client(Client):
```
andher1802, line 6: I did this naming, hahaha, I totally missed it. Thanks for fixing.

up42/asset.py
```diff
@@ -166,3 +173,57 @@ def download(
...
+            output_directory = (
+                Path.cwd()
+                / f"project_{self.auth.project_id}/asset_{self.asset_id}/{stac_asset.title}"
+            )
+        else:
+            output_directory = Path(output_directory)
+        output_directory.mkdir(parents=True, exist_ok=True)
+        logger.info(f"Download directory: {str(output_directory)}")
+        try:
+            up42_signed_token = self.auth._get_auth_token()
```
andher1802, line 207: Is it not possible to use the auth._request method. I think this would be better, instead to add headers and use our own request. Also, once we migrate to account authentication we will have to change it everywhere is used. Let me know what do you think.

up42/auth.py
```diff
@@ -153,6 +154,36 @@ def _endpoint(self) -> str:
         """Gets the endpoint."""
         return f"https://api.up42.{self.env}"
 
+    def _get_auth_token(self):
```
andher1802, line 157: Do we need this? I think we can obtain this from the auth class.

With these two lines, you make sure the auth is authenticated and get the token. That makes everything you do in this function. However, by using the auth._request you would avoid this.

:) 
 
auth_class._get_token()
return auth_class.token

andher1802 (changes requested):
Hi Naman, 
Very nice work, I have added some comments. I think if we use the request from the auth you would simplify everything. 

If not possible, at least the token function will be super simple just by getting the token from the auth object.

up42/auth.py
seedlit: Indeed!
Removing this method altogether. Thanks :)

up42/asset.py
seedlit: The asset.href endpoint returns the contents of the file.
That's why I decided to implement our own request, similar to how we do it in `download` [method](https://github.com/up42/up42-py/blob/320b11955d67ee985b079b4d82cc955b0ea328a0/up42/utils.py#L164).

Do you prefer to refactor the `auth._request` method instead?

andher1802: no it is OK, then lets use the token from the auth objects.

tests/fixtures/fixtures_asset.py
seedlit: ahh, let me apply black formatting.
We should also configure pre-commit for SDK. 
I will create a ticket in the backlog

PR #473, opened 2023-10-12
AT-2939 Activating poetry and pre-commit

andher1802:

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


seedlit:
Hey @andher1802 , can you pull changes from master?
Also, maybe let's update the [circleCI config as well to use poetry](https://github.com/up42/up42-py/blob/0bdcd155c9ec7d0f4c02c1250b231aadc66032f3/.circleci/config.yml#L55)?

dustgalactic (approved):

.circleci/config.yml
```diff
@@ -18,12 +18,13 @@ jobs:
       - run:
           name: Install requirements
           command: |
-            pip install -r requirements.txt
-            pip install -r requirements-viz.txt
-            pip install -r requirements-dev.txt
+            rm `which poetry`
+            curl -sSL https://install.python-poetry.org | python -
```
seedlit, line 22: suggestion: maybe let's use the base image `cimg/python:3.8`, which comes with poetry pre-installed?
See [here](https://github.com/up42/sdk-tests/blob/64736c0c9e9beb5bc3ca1d26e345e6a6f6d7e5c9/.circleci/config.yml#L5) for reference

seedlit: might also be a good opportunity to start using 3.11 as base image. But let's tackle that in another ticket as this might get out of scope (if some tests start failing)

.circleci/config.yml
```diff
@@ -52,12 +53,13 @@ jobs:
       - run:
           name: Install requirements
           command: |
-            pip install -r requirements.txt
-            pip install -r requirements-viz.txt
-            pip install -r requirements-dev.txt
+            rm `which poetry`
```
seedlit, line 56: same as above

docs/theme_override_home/.icons/up42-logo.svg
seedlit, line 1: this is so interesting that pre-commit make changes to the svg as well :D
Not sure, what changed here though

pyproject.toml
```diff
@@ -0,0 +1,53 @@
+[tool.poetry]
+name = "up42-py"
+version = "0.32.0"
+description = ""
```
seedlit, line 4: consider adding description, authors, and licence. 
You can take a look [here](https://github.com/up42/up42-py/blob/configure_poetry/pyproject.toml) for reference

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+authors = ["Your Name <you@example.com>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = ">=3.8,<3.12"
+requests =  "2.31.0"
```
seedlit, line 10: is there a reason we want to pin these dependencies to this specific version? If not, let's make it flexible by `"^2.31.0"`

Same for rest of the dependencies.

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+requests-oauthlib = "1.3.1"
+tenacity = "8.2.2"
+tqdm = "4.66.0"
+geojson = "3.0.1"
+geopandas = "0.13.2"
+pystac-client = "0.7.2"
+mkdocs-material = ">=8.5.11"
```
seedlit, line 17: let's try to stick to one convention.
Let's either use `">=8.5.11"` or `"^8.5.11"`, but not both :) 

My personal taste would be to use `"^8.5.11"`, since we use this in other python repos as well

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+mkdocs-autolinks-plugin = ">=0.6.0"
+mkdocs-macros-plugin = ">=0.7.0"
+rasterio = ">=1.3.8"
+folium = ">=0.14.0"
+branca = ">=0.6.0"
+matplotlib = ">=3.7.2"
+descartes = ">=1.1.0"
+cleo = "^2.0.1"
+
+[tool.poetry.dev-dependencies]
```
seedlit, line 31: we should also add the visualisation dependencies as another group, no?

.circleci/config.yml
```diff
@@ -18,12 +18,13 @@ jobs:
...
```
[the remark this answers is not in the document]
andher1802: Yes, there is an issue with the poetry version that comes with the python3.8 image. Therefore, I need to remove poetry and install the latest version to make the install run. https://medium.com/@life-is-short-so-enjoy-it/python-poetry-modulenotfounderror-no-module-named-cleo-1f924239b06c

.circleci/config.yml
andher1802: Yes, problem that I mentioned before. For the python 3.11 as base image, what we can do actually is test the SDK for different versions, this should work on all versions that we support.

.circleci/config.yml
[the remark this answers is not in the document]
seedlit: In that case I strongly recommend to use `cimg/python:3.8` base image, which comes with poetry pre-installed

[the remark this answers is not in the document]
andher1802: That was not the image that had the issue?

pyproject.toml
andher1802: This can be a good idea, but I thing >= is not the same as ^. I will keep the docs requirements as >= to be consistent with the pip (Hopefully after a couple of iterations we will be removing pip) and then we can deal with the requirements properly. 

The problem is that we have some versions in pip that uses the >= which means any version more than the specified. The ^ operator means that the requirement should be at least the specified version (it means that the version need to exist), and for the docs requirements we need to check which versions are the ones that exist and do not cause incompatibilities. I will add a new ticket to deal with versions in the proper way for poetry, in the mean whiles. This makes the trick to activate poetry and pre-commit which make easier to extend the SDK.

.circleci/config.yml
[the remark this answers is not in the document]
andher1802: Yes, this seems to do the trick. We will use the cimg/python:3.8 base image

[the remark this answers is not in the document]
andher1802: Thanks

andher1802:
> Nice! Left some comments :) Since we are adding poetry, I would assume we would remove the requirements.txt files as well? In that case, we also need to make changes to [setup.py](https://github.com/up42/up42-py/blob/AT-2939_SDK_Refactoring/setup.py).
> 
> Or, would you rather prefer to create a ticket for that and let's handle that in another PR? :)

About this, I think we need to give some time and plan how to remove pip and document this change. We can create a new ticket for this.

pyproject.toml
seedlit: yeah, `>=` seems to be better choice.
[https://python-poetry.org/docs/dependency-specification/#caret-requirementssion ](https://python-poetry.org/docs/dependency-specification/#caret-requirementssion)

> The ^ operator means that the requirement should be at least the specified version (it means that the version need to exist)

But that is also true for `>=`, no?

.circleci/config.yml
```diff
@@ -10,16 +10,14 @@ jobs:
...
     steps:
       - attach_workspace:
           at: *workspace-dir
       - checkout
       - run:
           name: Install requirements
           command: |
-            rm `which poetry`
-            curl -sSL https://install.python-poetry.org | python -
             poetry lock
```
seedlit, line 21: what does this command do?
Usually, `poetry install --no-ansi` does the trick

pyproject.toml
```diff
@@ -1,52 +1,57 @@
...
 [tool.poetry.dependencies]
 python = ">=3.8,<3.12"
-requests =  "2.31.0"
-requests-oauthlib = "1.3.1"
-tenacity = "8.2.2"
-tqdm = "4.66.0"
-geojson = "3.0.1"
-geopandas = "0.13.2"
-pystac-client = "0.7.2"
+requests =  "^2.31.0"
```
seedlit, line 11: nitpick: how about we use `>=` everywhere?

seedlit (commented):
left a couple more nitpick comments. Sorry.

.circleci/config.yml
seedlit: I think this command will create a lock file based on `pyproject.toml` file. But since we already have a `poetry.lock file`, we don't need this command.
[https://python-poetry.org/docs/cli/#lock](https://python-poetry.org/docs/cli/#lock)

seedlit: interesting. 
Can you point me to the failing workflow?

pyproject.toml
andher1802: Honestly, I do not think that the issue is that we have to use the same everywhere as they do not serve the same purpose. 

>= (Greater Than or Equal To):
It's used to specify a minimum version requirement for a dependency.
For example, if you specify "requests >=2.0.0", it means your project requires at least version 2.0.0 of the "requests" library, but it can use any version that is greater or equal to that.

^ (Caret):
It's used to specify a compatible version range that allows for updates within a major version.
For instance, if you specify "requests ^2.0.0", it means your project can use any version within the 2.x range. It allows minor and patch version updates but keeps the major version fixed.
The main difference between the two is in how they handle version updates:

>= is more strict and only allows for patch and minor version updates.
^ is more permissive and allows for updates up to, but not including, the next major version.

For example, if the latest version of a package is 2.1.0:
"requests >=2.0.0" would accept any version from 2.0.0 to 2.1.0 but not 2.2.0.
"requests ^2.0.0" would accept any version from 2.0.0 to 3.0.0 but not 3.0.0 or higher.

So, you would use >= if you want to set a minimum version requirement and be more conservative about updates. You would use ^ if you want to allow for some flexibility and automatic updates within the same major version.

Unfortunately, in our case we need to figure out which dependencies can create conflicts and this can take some time, specially because when we deal with dependencies here we are dealing with not only pip but also conda releases. I think in the SDK case the dependencies are more complex and will need a separate effort to tackle, for Now I left the exact same that we have in the pip. Once we get rid of the requirement.txt files we can make sure all dependencies are working together.

WDYT.

.circleci/config.yml
andher1802: sure I think is this one https://app.circleci.com/pipelines/github/up42/up42-py/3245/workflows/7df060c0-19d3-4267-a018-066fa99f97ef/jobs/3661

pyproject.toml
andher1802: Thanks, man, not that I don't want to add it. I agree with you, but I think this dependency issue requires more careful consideration. Every time I touch the requirements, I am getting some error. And I not to mention when the time of release comes, we will have issues with the conda release that has a separate configuration for dependencies. I am still not sure which dependencies conda we will conflict. 

I really appreciate that you give detail review to this PRs. 

I am preparing myself for the release time. It will be nasty. :)

seedlit (approved):

PR #475, opened 2023-10-13
[AT-3079] Expose STAC asset signed url

seedlit:
Jira ticket[ here](https://up42.atlassian.net/browse/AT-3079).
This PR adds method to get signed URL for a STAC asset.

Minimalistic example to generate a signed URL for a STAC Asset:

`import up42`

`up42.authenticate(
    project_api_key="your_api_key",
    project_id="your_project_id",
    env="dev",
)`
`up42_asset = up42.initialize_asset(asset_id="your_asset_id")`
`stac_items = up42_asset.stac_items.items`
`target_stac_item = stac_items[0]`
`stac_assets = target_stac_item.assets`
`liscense_asset = stac_assets.get("licenseagreement-1.1.txt")`
`liscence_url = up42_asset.get_stac_asset_url(liscense_asset)`
`print(liscence_url)`


up42/asset.py
```diff
@@ -128,12 +126,28 @@ def update_metadata(
...
+            url = f"{self.auth._endpoint()}/v2/assets/{self.asset_id}/download-url"
+        else:
+            url = f"{self.auth._endpoint()}/v2/assets/{stac_asset_id}/download-url"
+        response_json = self.auth._request(request_type=request_type, url=url)
         download_url = response_json["url"]
         return download_url
 
+    def get_stac_asset_url(self, stac_asset: pystac.Asset):
+        """
+        Returns the signed URL for the STAC Asset.
```
saad-up42, line 140: As it has a 30 second validity period, this would be a pre-signed URL instead of just a signed URL.

up42/asset.py
```diff
@@ -128,12 +126,28 @@ def update_metadata(
...
+        Args:
+            stac_asset: pystac Asset object.
+
+        Returns:
+            Signed URL for the STAC Asset.
```
saad-up42, line 145: I believe this should be a pre-signed URL.

up42/asset.py
saad-up42, line 140:
```suggestion
        Returns the pre-signed URL for the STAC Asset.
```
As the URL has a 30 second validity, this would be a pre-signed URL.

up42/asset.py
saad-up42, line 145:
```suggestion
            Pre-signed URL for the STAC Asset.
```
I believe this is technically a pre-signed URL.

saad-up42:
### Changes introduced by @saad-up42:

- Add `get_stac_asset_url()`  function to asset-reference.md
- Update asset-example.ipynb to include new function
- Update structure.md to include new function
- Add new entry to CHANGELOG.md 
- Update main.html to update the banner text
- Update asset.py to use "pre-signed URL" term (x2)

### Relevant tickets

- Documentation JIRA task [DOC-783](https://up42.atlassian.net/browse/DOC-783)
- Analytics JIRA task [AT-3079](https://up42.atlassian.net/browse/AT-3079)
 
 CC: @seedlit @andher1802 @dustgalactic

andher1802 (approved):
LGTM

dustgalactic (approved):

up42/asset.py
seedlit: @saad-up42 maybe I am wrong, but isn't the term "pre-signed URL" very specific to AWS S3 buckets?
If not, what's the difference between signed URL and pre-signed URL? :)

dustgalactic: @saad-up42 Where did you read about this distinction (limited validity = pre-signed, other = signed)? Because e.g. [here](https://cloud.google.com/storage/docs/access-control/signed-urls) it says "Anyone who knows the _signed_ URL can access the resource until the expiration time for the URL is reached or the key used to sign the URL is rotated.", so seems like Naman is right that it would be considered a signed URL

saad-up42: @dustgalactic The [reference](https://stackoverflow.com/questions/20862195/what-is-difference-between-pre-signed-url-and-signed-url) I used also refers to it for AWS S3 buckets, so @seedlit may be right and it might just be a signed URL then.

If that's the case, we can change it back to a signed URL here and in other instances like the API reference.

dustgalactic: @saad-up42 As for the API reference, don't do it right now, I'll combine it with another task related to the asset repo. 

As for the SDK reference, @seedlit it's your call, we can fix it [in this PR](https://github.com/up42/up42-py/pull/479) we're working on right now or leave it up to you

saad-up42: > yeah, let's fix it in the PR you are working on right now :)

@dustgalactic if that's okay I will include the changes in #479 

API changes will need to be made elsewhere.

PR #486, opened 2023-10-25
AT-3835 Account based auth class

andher1802:
This PR includes all the changes to migrate to account-based authentication, while we still allow project-based authentication for some time. 

This PR is tracked here in [AT-3556](https://up42.atlassian.net/jira/software/projects/AT/boards/63?assignee=61d8425a49f195006962b9a3&selectedIssue=AT-3556)

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

[AT-3556]: https://up42.atlassian.net/browse/AT-3556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.gitignore
```diff
@@ -145,3 +145,6 @@ docs/mkdocs_macros
 project_*/
 
 .DS_Store
+
```
andher1802, line 148: temporal change, please ignore it

up42/auth.py
```diff
@@ -65,11 +71,23 @@ def is_http_429_error(exception):
...
+        },
+    }
+
     def __init__(
         self,
+        auth_type: str = AuthType.PROJECT.value,
         cfg_file: Union[str, Path, None] = None,
-        project_id: Optional[str] = None,
-        project_api_key: Optional[str] = None,
+        credentials_id: Optional[str] = None,
```
seedlit, line 89: ahh, so users will have use these keywords now.
Hmm, this will force them to make changes to their pipeline.
Since we will be deprecating it in future anyway, I would suggest to keep the existing keys.

up42/auth.py
```diff
@@ -100,6 +119,11 @@ def __init__(
         except KeyError:
             self.authenticate = True
 
+        if all([arg in kwargs for arg in ["project_id", "project_api_key"]]):
+            self.credentials_id = kwargs["project_id"]
```
seedlit, line 123: okay I see it.
The old keys would still work 👍🏻

up42/auth.py
```diff
@@ -100,6 +119,11 @@ def __init__(
...
+            self.credentials_key = kwargs["project_api_key"]
+            self.project_id = self.credentials_id
```
seedlit, line 125: this would fail if it's account based authentication right?

up42/auth.py
```diff
@@ -115,26 +139,29 @@ def _find_credentials(self) -> None:
...
-                raise ValueError("Provide project_id and project_api_key via arguments or config file!")
+                raise ValueError("Provide credentials via arguments or config file!")
 
             # Source credentials from config file.
             try:
-                with open(self.cfg_file) as src:
+                if self.auth_type == AuthType.PROJECT.value:
+                    logger.info(
+                        "Project based authentication will be deprecated soon."
+                        "Please follow the [authentication guidelines](https://sdk.up42.com/authentication/)."
```
seedlit, line 151: Nice!
I guess @up42/documentation will update this page in parallel when we release this?

up42/auth.py
```diff
@@ -115,26 +139,29 @@ def _find_credentials(self) -> None:
...
             if self.cfg_file is None:
-                raise ValueError("Provide project_id and project_api_key via arguments or config file!")
+                raise ValueError("Provide credentials via arguments or config file!")
 
             # Source credentials from config file.
             try:
-                with open(self.cfg_file) as src:
+                if self.auth_type == AuthType.PROJECT.value:
+                    logger.info(
+                        "Project based authentication will be deprecated soon."
```
seedlit, line 150: We can actually add a date here as well.
Maybe something like `Project based authentication will be deprecated in March 2024. Please use Account based authentication instead.`

up42/auth.py
```diff
@@ -115,26 +139,29 @@ def _find_credentials(self) -> None:
...
                     try:
-                        self.project_id = config["project_id"]
-                        self.project_api_key = config["project_api_key"]
+                        self.credentials_id = config[self.cred_mapping[self.auth_type]["credentials_id"]]
+                        self.credentials_key = config[self.cred_mapping[self.auth_type]["credentials_key"]]
                     except KeyError as e:
-                        raise ValueError(
-                            "Provided config file does not contain project_id and " "project_api_key!"
-                        ) from e
+                        raise ValueError("Provided config file does not contain valid credentials.") from e
```
seedlit, line 159: We actually don't know here whether the credentials are valid or not.
All we know is the keys are wrong in the config file.
Consider re-phrasing the ValueError string.

up42/auth.py
```diff
@@ -144,10 +171,39 @@ def _endpoint(self) -> str:
...
+            req_body = {
+                "grant_type": "password",
+                "username": self.credentials_id,
+                "password": self.credentials_key,
+            }
+            token_response = requests.post(
+                url=self._endpoint() + "/oauth/token",
+                data=req_body,
+                headers=req_headers,
+                timeout=1200,
```
seedlit, line 194: 20 mins seem like a very long time to wait before we raise Timeout exception. Consider lowering it something like 120 seconds.

Also, consider defining this as a global variable and passing it here. Something like:
`TIMEOUT_DURATION = 120`
`timeout = TIMEOUT_DURATION`

up42/auth.py
```diff
@@ -144,10 +171,39 @@ def _endpoint(self) -> str:
...
+            )
+            if token_response.status_code != 200:
+                raise ValueError("Authentication was not successful, check the provided credentials.")
+        except requests.exceptions.RequestException as err:
```
seedlit, line 198: we are raising the same error message twice.
Can we combine the two?

up42/auth.py
```diff
@@ -156,10 +212,10 @@ def _get_token(self):
         self.token = token_response["data"]["accessToken"]
 
     def _get_workspace(self) -> None:
-        """Get workspace id belonging to authenticated project."""
-        url = f"https://api.up42.{self.env}/projects/{self.project_id}"
+        """Get user id belonging to authenticated account."""
```
seedlit, line 215: if this is not returning workspace id anymore, consider renaming the function.

up42/auth.py
```diff
@@ -204,6 +260,7 @@ def _request_helper(
                 url=url,
                 data=json.dumps(data),
                 headers=headers,
+                timeout=1200,
```
seedlit, line 263: same comment on timeout as above :)

seedlit (commented):
left a few comments :)

up42/auth.py
andher1802: Yes, I added support for the previous ones. :)

up42/auth.py
andher1802: Yes, just FYI workspace_id and user_id are the same value. I will change it.

up42/auth.py
andher1802: Not combine but raise more appropriate errors. First was for a not valid credentials the later is for network errors.

seedlit: of-course! 🤦🏻

up42/auth.py
andher1802: Yes actually this is the new doc page, @dustgalactic pointed me to that link

dustgalactic: https://up42.atlassian.net/browse/DOC-789

up42/workflow.py
```diff
@@ -44,12 +44,12 @@ class Workflow:
...
         self,
         auth: Auth,
-        project_id: str,
         workflow_id: str,
+        project_id: Optional[str] = None,
         workflow_info: Optional[dict] = None,
     ):
         self.auth = auth
-        self.project_id = project_id
+        self.project_id = project_id if project_id is not None else self.auth.credentials_id
```
andher1802, line 52: I know this can be the opposite of your idea to add the project_id but if this is not included it would mean a breaking change on the client side. Please, correct me if I am wrong.

andher1802: Same apply to other classes.

up42/estimation.py
```diff
@@ -29,7 +29,7 @@ def __init__(
                 },]
         """
         self.auth = auth
-        self.project_id = project_id
+        self.project_id = project_id if project_id is not None else self.auth.credentials_id
```
andher1802, line 32: I know this can be the opposite of your idea to add the project_id but if this is not included it would mean a breaking change on the client side. Please, correct me if I am wrong.

README-dev.md
```diff
@@ -18,13 +18,13 @@ cd up42-py
...
-3. Create a new project on [UP42](https://up42.com).
+3. Follow authentication guidelines [Authenticate](/docs/authentication.md).
 
-4. Create a `config.json` file and fill in the [project credentials](https://docs.up42.com/developers/authentication#step-1-find-project-credentials).
+4. Create a `config.json` file and fill in the credentials.
 ```json
 {
-  "project_id": "...",
-  "project_api_key": "..."
+  "credentials_id": "...",
```
dustgalactic, line 26: We already provide config.json instructions in `authenticate.md`. This is how I previously edited the file: [link](https://github.com/up42/up42-py/pull/486/commits/09dae048100fbc5c8658b962840861eec68a10e3#diff-faeb6913d3584750b35461ea79794816d906655f7af245cdd934e118271e7cb1R21).

up42/auth.py
```diff
@@ -100,41 +123,55 @@ def __init__(
         except KeyError:
             self.authenticate = True
 
+        if all([arg in kwargs for arg in ["project_id", "project_api_key"]]) and auth_type == AuthType.PROJECT.value:
+            self.credentials_id = kwargs["project_id"]
+            self.credentials_key = kwargs["project_api_key"]
+            self.project_id = self.credentials_id
```
javidq, line 129: self.kwargs.get("project_id")

up42/auth.py
```diff
@@ -100,41 +123,55 @@ def __init__(
         except KeyError:
             self.authenticate = True
 
+        if all([arg in kwargs for arg in ["project_id", "project_api_key"]]) and auth_type == AuthType.PROJECT.value:
+            self.credentials_id = kwargs["project_id"]
```
javidq, line 127: should be private and the same goes for key

up42/auth.py
```diff
@@ -100,41 +123,55 @@ def __init__(
...
+            self.project_id = self.credentials_id
+
         if self.authenticate:
             self._find_credentials()
             self._get_token()
-            self._get_workspace()
+            self._get_user_id()
             logger.info("Authentication with UP42 successful!")
 
     def __repr__(self):
```
javidq, line 137: Not needed at all

up42/auth.py
```diff
@@ -100,41 +123,55 @@ def __init__(
...
         env_string = f" ,{self.env}" if self.env != "com" else ""
-        return f"UP42ProjectAuth(project_id={self.project_id}{env_string})"
+        if self.auth_type == AuthType.ACCOUNT.value:
+            return f"UP42UserAuth(user_id={self.credentials_id}{env_string})"
+        return f"UP42ProjectAuth(project_id={self.credentials_id}{env_string})"
 
     def _find_credentials(self) -> None:
```
javidq, line 143: We need to split this into 2 methods. One will choose the credentials source between config file and arguments, something like:
```python
    def _choose_credential_source(cfg_file: str, kwargs: dict):
        config = {}
        try:
            with open(cfg_file, encoding="utf-8") as src:
                config.update(json.load(src))
            logger.info("Got credentials from config file.")
        except FileNotFoundError as e:
            raise ValueError("Selected config file does not exist!") from e
        if set(config.keys()).intersection(set(kwargs.keys())):
            raise ValueError("Credentials must be provided either via config file or arguments, but not both")
        return config if config else kwargs
```
The second will read the parameters from the chosen source, smith like
```python
    def _set_credentials(self, source: dict):
        schemas = {
            "user": ("username", "password"),
            "project": ("project_id", "project_api_key")
        }
        token_retrievers = {"project": self._get_project_token, "user": self._get_account_token}

        for parameters, schema in schemas.items():
            if set(parameters).issubset(source.keys()):
                self._credentials_id = source[parameters[0]]
                self._credentials_key = source[parameters[1]]
                self._get_token = token_retrievers[schema]
        if self._credentials_id is None:
            # raise error
```

javidq (changes requested):

up42/auth.py
andher1802: The way it works currently for project-based authentication is: 

If there are both arguments and cfg_file, the cfg_file has the priority and we log that we choose the cfg_file over the parameters. 

My question is: 
1. Now we have 4 options. cfg_project_based, keys_project_based, cfg_account_based, keys_account_based. 
2. Should we follow that cfg_based are priority.
3. Also, I think that we should also allow project_based and account_based at the same time provided, but let the account based to have priority. 
4. The only case we raise an error should be with not complete credentials provided.

andher1802: For consideration, what about something like this to add some hierachy if the user adds project-based or account-based via cfg_files and arguments.
`
    def _choose_credentials(username, password, project_id, project_api_key, config_file) -> tuple[str, str]:
        if config_file:
            try:
                with open(config_file, 'r') as file:
                    config_data = json.load(file)
                    if 'project_id' in config_data and 'project_api_key' in config_data:
                        return config_data['project_id'], config_data['project_api_key'] 
                    if 'username' in config_data and 'password' in config_data:
                        return config_data['username'], config_data['password'] 
            except FileNotFoundError as e:
                raise ValueError("Selected config file does not exist!") from e

        if project_id and project_api_key:
            return project_id, project_api_key
        
        if username and password:
            return username, password 
        raise ValueError("No credentials provided either via config file or arguments.")

`

up42/auth.py
andher1802: Please take a look to the alternative approach. Following your same logic but only set `self.project_id` if provided via arguments or cfg_file. Maybe, we need to add some further checking in the initialize as you suggest.

.pre-commit-config.yaml
```diff
@@ -8,7 +8,7 @@ repos:
       - id: end-of-file-fixer
         exclude: ^up42/_version\.txt$|.github/CODEOWNERS
       - id: check-added-large-files
-        args: ["--maxkb=1000"]
+        args: ["--maxkb=3000"]
```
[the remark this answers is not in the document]
javidq: Unaddressed

up42/auth.py
```diff
@@ -100,66 +111,112 @@ def __init__(
...
+        schemas = {
+            "user": ("username", "password"),
+            "project": ("project_id", "project_api_key"),
+        }
+        token_retrievers = {
+            "user": self._get_token_account_based,
+            "project": self._get_token_project_based,
+        }
+
+        if set(source.keys()) == {"project_id"}:
```
javidq, line 155: Why is this needed?

up42/auth.py
```diff
@@ -100,66 +111,112 @@ def __init__(
...
+            return
+
+        for parameters, schema in schemas.items():
```
javidq, line 158: should be reverse

up42/auth.py
javidq: Unaddressed

up42/auth.py
```diff
@@ -100,66 +111,112 @@ def __init__(
         except KeyError:
             self.authenticate = True
 
+        self._credentials_id = None
```
javidq, line 114: I'd propose to move this to set_credentials

up42/auth.py
```diff
@@ -70,6 +75,8 @@ def __init__(
         cfg_file: Union[str, Path, None] = None,
         project_id: Optional[str] = None,
         project_api_key: Optional[str] = None,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
```
javidq, line 79: not needed, let them come from kwargs

up42/auth.py
```diff
@@ -86,11 +93,15 @@ def __init__(
             project_api_key: The project-specific API key.
         """
         self.cfg_file = cfg_file
```
javidq, line 95: not needed

up42/auth.py
```diff
@@ -86,11 +93,15 @@ def __init__(
             project_api_key: The project-specific API key.
```
javidq, line 93: Doc string to be updated

up42/initialization.py
```diff
@@ -36,7 +36,11 @@
 
 def _check_deprecated_implicit_project_id(project_id: Optional[str]):
     if not project_id:
-        warn("Provide the project ID as the value of the `project_id` argument.", DeprecationWarning, stacklevel=2)
```
javidq, line 39: This does not come from my PR, where's the PR which changed this?

up42/main.py
```diff
@@ -38,6 +38,8 @@ def authenticate(
     cfg_file: Union[str, Path] = None,
     project_id: Optional[str] = None,
     project_api_key: Optional[str] = None,
+    username: Optional[str] = None,
```
javidq, line 41: docstring to be updated

up42/project.py
```diff
@@ -16,10 +16,6 @@ class Project:
     new workflows, query already existing workflows & jobs in the project and manage
     the project settings.
 
-    Create a new project on the
-    [**UP42 Console website**](https://sdk.up42.com/authentication/#get-your-project-credentials).
-
-    Use an existing project:
     ```python
     up42.authenticate(project_id="uz92-8uo0-4dc9-ab1d-06d7ec1a5321",
```
javidq, line 20: Why not to drop this line as well? Also in __init__.py docstring

up42/initialization.py
```diff
@@ -95,7 +99,11 @@ def initialize_job(job_id: str, project_id: Optional[str] = None) -> "Job":
         project_id: The id of the UP42 project, containing the job
     """
     _check_deprecated_implicit_project_id(project_id)
-    job = Job(auth=main._auth, job_id=job_id, project_id=project_id or str(main._auth.project_id))
+    job = Job(
+        auth=main._auth,
+        job_id=job_id,
+        project_id=project_id or str(main._auth.project_id),
```
javidq, line 105: This is not what I proposed. I meant smith like this:
```python
def _get_project_id(project_id: Optional[str]):
    if not project_id:
        warn(
            "Provide the project ID as the value of the `project_id` argument.",
            DeprecationWarning,
            stacklevel=2,
        )
    result = project_id or str(main._auth.project_id)
    if not result:
        raise ValueError("Project ID can't be null")
    return result
```
and use it everywhere instead of `project_id or str(main._auth.project_id)` combination in all 5 cases

javidq (changes requested):

tests/test_auth.py
```diff
@@ -53,12 +62,35 @@ def test_get_token(auth_mock):
     assert auth_mock.token == TOKEN
 
 
+def test_repr_project():
```
javidq, line 65: None of these are needed, auth an internally used classed, should not be printed anywhere anyway

tests/test_auth.py
```diff
@@ -22,23 +31,23 @@ def test_auth_kwargs():
...
-    auth_mock.cfg_file = fp
 
-    auth_mock._find_credentials()
-    assert auth_mock.project_id is not None
-    assert auth_mock.project_api_key is not None
+def test_find_credentials_cfg_file():
+    fp = Path(__file__).resolve().parent / "mock_data" / "test_config.json"
+    auth = Auth(cfg_file=fp, authenticate=False)
+    assert auth._credentials_id is not None
+    assert auth._credentials_key is not None
```
javidq, line 51: We're missing lots of tests, but most importantly the test of the new authentication method

javidq (changes requested):

up42/auth.py
andher1802: Because of this use case. The user Initializes the project by only providing the project_id (not the project_api_key).
The suggested workflow in set_credentials is that if there is not pair (username, password) or (project_id, project_api_key) it raises an error. Therefore, by adding the case of only adding project_id and return before authentication, makes this use case available.

```
def test_should_initialize_project_with_project_id(auth_mock, requests_mock):
    up42.authenticate(
        project_id=PROJECT_ID,
        authenticate=False,
    )
```

up42/auth.py
andher1802: call a dict items in python is like this

`for key, value in dict.items():`

The order is key, value, the suggested logic worked with the value compares as sub set of the source. I think the problem is the order in the for not naming. I will fix it. 👍

up42/initialization.py
andher1802: Your PR was merged. What changed is that the pre-commit hook on my end breaks down single lines into multiple lines. It is a recurrent change on my end automatically when I run the pre-commit hook.

andher1802: This is what you changed in your PR. same thing but single line. Not sure if single line - multi line issue is too relevant, but I think multiline makes it more readable, however no strong preference on my side.

Screenshot 2023-11-02 at 08 56 25

tests/test_job.py
```diff
@@ -177,12 +200,14 @@ def test_job_download_result_dimap_live(auth_live, project_id_live):
         assert Path(out_files[20]).exists()
         assert Path(out_files[-1]).exists()
         assert "data.json" in [Path(of).name for of in out_files]
-        assert len(out_files) == 46
+        assert len(out_files) == 44
```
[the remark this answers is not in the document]
andher1802: Outdated live test. Usually this is not caught because live test should be triggered manually. Just making sure that all test works .

tests/fixtures/fixtures_project.py
```diff
@@ -52,7 +57,9 @@ def project_mock(auth_mock, requests_mock):
     requests_mock.get(url=url_workflow_info, json=json_workflow_info)
 
     # get_workflows
-    url_get_workflows = f"{project.auth._endpoint()}/projects/" f"{project.project_id}/workflows"
+    url_get_workflows = (
+        f"{project.auth._endpoint()}/projects/" f"{project.project_id}/workflows"
```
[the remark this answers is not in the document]
andher1802: automatic change from my side. I am not sure if this is pre-commit or linter related.

up42/auth.py
andher1802: I am not sure about this comment. There are still some endpoints that require to add workspace_id or user_id (e.g. order endpoints V1, which will be soon migrated). I think that we still need workspace_id while we finish to migrate V1 endpoints. Let me know, what do you mean by not needed at all.

up42/project.py
andher1802: Added the suggestion from Daria, pointing to the authentication guidelines.

tests/fixtures/fixtures_auth.py
```diff
@@ -23,7 +23,16 @@ def auth_mock(requests_mock):
         url=url_get_workspace,
         json=json_get_workspace,
     )
-    auth = Auth(project_id=PROJECT_ID, project_api_key=PROJECT_APIKEY, authenticate=True)
+
+    url_get_workspace_v2 = f"https://api.up42.com/users/me"
```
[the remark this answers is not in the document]
andher1802: I think when you say the first one, you mean about why do we need the workspace_id. This is needed to make some endpoints from orders V1 work while we migrate. Then to make the mock auth work as well we need to mock the same endpoint. Please let me know if that is what you meant.

tests/test_auth.py
andher1802: Yes, we agreed to first change the auth class and then test.

tests/fixtures/fixtures_auth.py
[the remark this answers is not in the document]
andher1802: We don't, there were for old tests that don't exist anymore.

tests/test_auth.py
andher1802: I have added some initial fixtures and tests for the account_based authentication.

javidq: Let's avoid such practice in the future:
1. It creates a risk of adding untestable code which will be changed a follow PR anyways
2. It requires certain mental gymnastics to have in check that all changes from an earlier PR are tested in a follow up PR
3. It forces the creation of feature branches instead of merging to the main one in all cases

tests/fixtures/fixtures_project.py
```diff
@@ -149,9 +162,12 @@ def project_live(auth_live, project_id_live):
 def project_mock_max_concurrent_jobs(project_mock):
     def _project_mock_max_concurrent_jobs(maximum=5):
         m = requests_mock.Mocker()
-        url_project_info = f"{project_mock.auth._endpoint()}/projects/{project_mock.project_id}"
+        url_project_info = f"{project_mock.auth._endpoint()}/projects/{PROJECT_ID}"
         m.get(url=url_project_info, json={"data": {"xyz": 789}, "error": {}})
-        url_project_settings = f"{project_mock.auth._endpoint()}/projects" f"/{project_mock.project_id}/settings"
+        url_project_settings = (
+            f"{project_mock.auth._endpoint()}/projects"
+            f"/{project_mock.project_id}/settings"
```
javidq, line 169: Isn't it PROJECT_ID?

tests/test_estimation.py
```diff
@@ -2,7 +2,15 @@
...
 # pylint: disable=unused-import,wrong-import-order
 from .context import Estimation
-from .fixtures import JSON_WORKFLOW_ESTIMATION, PROJECT_ID, auth_live, auth_mock, estimation_mock
+from .fixtures import (
+    JSON_WORKFLOW_ESTIMATION,
+    PROJECT_ID,
+    auth_live,
+    auth_mock,
+    estimation_mock,
+    project_api_key_live,
```
javidq, line 11: Unused

tests/fixtures/fixtures_auth.py
```diff
@@ -42,6 +44,27 @@ def auth_mock(requests_mock):
     return auth
 
 
+@pytest.fixture()
```
javidq, line 47: () is not needed

tests/fixtures/fixtures_auth.py
```diff
@@ -42,6 +44,27 @@ def auth_mock(requests_mock):
...
+        "token_type": "bearer",
+    }
+    requests_mock.post(url_get_token, json=json_get_token)
+    url_get_workspace = f"https://api.up42.com/users/me"
+    json_get_workspace = {"data": {"id": WORKSPACE_ID}}
+    requests_mock.get(
+        url=url_get_workspace,
+        json=json_get_workspace,
+    )
+    auth = Auth(username=USERNAME_TEST, password=PASSWORD_TEST, authenticate=True)
```
javidq, line 64: can be inlined

tests/fixtures/fixtures_auth.py
```diff
@@ -42,6 +44,27 @@ def auth_mock(requests_mock):
     return auth
 
 
+@pytest.fixture()
+def auth_account_mock(requests_mock):
+    USERNAME_TEST = "something@up42.com"
```
javidq, line 49: It's someone)) Let's change it to `user@up42.com`

javidq: And these two are local variables, why are they named as module constants?

tests/fixtures/fixtures_auth.py
```diff
@@ -42,6 +44,27 @@ def auth_mock(requests_mock):
     return auth
 
 
+@pytest.fixture()
+def auth_account_mock(requests_mock):
```
javidq, line 48: We're settings credit balance and blocks above in project case, why don't we need them here?

tests/test_auth.py
```diff
@@ -48,32 +61,38 @@ def test_endpoint(auth_mock):
     assert auth_mock._endpoint() == "https://api.up42.abc"
 
 
-def test_get_token(auth_mock):
-    auth_mock._get_token()
-    assert auth_mock.token == TOKEN
+@pytest.mark.parametrize("authentication_fixture", ("auth_mock", "auth_account_mock"))
```
javidq, line 64: Actually we need to change in auth_fixtures to 
```
@pytest.fixture(params=["project", "account"])
def auth(request, auth_project_mock, auth_account_mock):
    mocks = {"project": auth_project_mock, "account": auth_account_mock}
    return mocks[request.param]
```
to run all tests with both mocks if they don't specify which kind of auth type they need

javidq: And the same below

tests/test_auth.py
```diff
@@ -48,32 +61,38 @@ def test_endpoint(auth_mock):
...
-def test_get_token_live(auth_live):
-    assert hasattr(auth_live, "token")
+@pytest.mark.parametrize("authentication_fixture", ("auth_live", "auth_account_live"))
+def test_get_token_live(request, authentication_fixture):
+    assert hasattr(request.getfixturevalue(authentication_fixture), "token")
 
 
 def test_get_workspace(auth_mock):
-    auth_mock._get_workspace()
+    auth_mock._get_user_id()
```
javidq, line 88: I'm still worried about this being a breaking change for existing pipelines

javidq (changes requested):

up42/auth.py
andher1802: clear.

tests/fixtures/fixtures_auth.py
andher1802: Because account based authentication is not used for the other tests. If we add to them, we would need to add those mock responses.

up42/initialization.py
andher1802: This comes from a PR made to the parent branch by @dustgalactic. Take into account that she is involved in all the logs and communication to the users. I revert the change but JFYI. :)

tests/fixtures/fixtures_auth.py
andher1802: Actually this was not required by the new testing approach of using both auth_mocks.

examples/data-block-examples/exactearth-hvt.ipynb
```diff
@@ -38,7 +38,7 @@
    "source": [
     "up42.authenticate(cfg_file=\"config.json\")\n",
     "up42.settings(log=False)\n",
-    "project = up42.initialize_project()"
+    "project = up42.initialize_project(project_id=\"68567134-27ad-7bd7-4b65-d61adb11fc78\")"
```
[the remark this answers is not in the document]
saad-up42: Our documentation philosophy is to make the instructions self-sufficient where possible, and when we have two competing methods to focus on the simpler one for a user to achieve. 

With this in mind, I would change this to use the same approach as the other notebooks.

@javidq

javidq (approved):

PR #497, opened 2023-11-21
add deprecation warnings

seedlit:
Jira ticket [here](https://up42.atlassian.net/browse/AT-3595).
@up42/documentation @andher1802 please take a look and let me know if you have any suggestions regarding the warning message.

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/initialization.py
```diff
@@ -54,6 +54,7 @@ def initialize_project(project_id: Optional[str] = None) -> "Project":
     Args:
         project_id: The UP42 project id
     """
+    warn("Projects will be deprecated soon.")
```
dustgalactic, line 57: Here and below:
* I'm not sure about "soon". The functionality is either active or deprecated (meaning = will be removed in the nearest future). So I'd go with "X are deprecated".
* We can reuse the text I will be putting up as warnings on relevant doc pages:
"X are deprecated. The current analytics platform will be discontinued after January 31, 2024, and will be replaced by new processing functionalities." @seedlit

andher1802: I think we have a date right? Q1 2024 according to @kodamaaiko

dustgalactic: @seedlit Sounds good to me!

dustgalactic (approved):

up42/initialization.py
```diff
@@ -54,6 +54,10 @@ def initialize_project(project_id: Optional[str] = None) -> "Project":
     Args:
         project_id: The UP42 project id
     """
+    warn(
```
andher1802, line 57: I think we should add the DeprecationWarning class as this is part of the same group of warnings (Applies to the other logs).

The stacklevel would be 1 by default, it means that the message will be emitted from the location where the `warn` was called. Do we want to emmit the warning from the method itself or from the Class?.

andher1802 (commented):
Does this change need to be aligned with the documentation in the docs? 

I would say that logs should be also aligned with deprecations warnings in the documentation.

up42/initialization.py
seedlit: ah totally forgot.
Will update.
Thanks!

seedlit: For initialization functions, the stack_level should be 1.
So I would skip the param `stacklevel=2`

andher1802 (approved):
LGTM

PR #501, opened 2023-11-24
[AT-2924] configure poetry

seedlit:
Jira ticket [here.](https://up42.atlassian.net/browse/AT-2924)

Scope:
 - Removed setup.py and all requirements.txt file
 - Removed twine as dependency (instead, using poetry to publish)
 - updated CircleCI context
 - use poetry to publish docs
 - upgraded base image to python 3.11

NOTE: Tested that publication works by [publishing on test.pypi.org](https://test.pypi.org/project/up42-py-test/). I created a new account. Credentials can be found in 1Password --> `test.pypi.org credentials`

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


andher1802 (commented):
Should we remove _version.txt file from up42 folder?

andher1802 (commented):
When you publish, can you make sure that reading the sample geometries is still working? 

I mean, the below configuration in `setup.py` made sure that the `aoi_berlin.geojson` and `aoi_washington.geojson` were copied in the published version. We also had to add this in conda. I am not sure if by default these are added in poetry. (_version.txt should be removed)

```
package_data={"": ["_version.txt", "data/aoi_berlin.geojson", "data/aoi_washington.geojson"]}
```

andher1802 (commented):
Also: `packages=find_packages(exclude=("tests", "docs", "examples"))`
I think we removed these folders from the published version.

docs/README.md
```diff
@@ -8,7 +8,7 @@ Redirected to custom domain [https://sdk.up42.com/](https://sdk.up42.com/)
 ## Installation for doc development:
 ```
 cd docs
-pip install -r requirements_dev.txt
+poetry install only docs
```
javidq, line 11: It's confusing that we have poetry install here and make install in dev readme, we need a unified approach

Makefile
```diff
@@ -5,14 +5,10 @@ env:
 	workon up42-py
 
 install:
```
javidq, line 7: 1. pytest - to switch to poetry
2. e2e - to switch to poetry. Do we actually need this as a separate step?
3. serve - better to switch to poetry to make sure that it runs with the specified dependencies
4. gh-pages - poetry

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "up42-py"
 version = "0.33.1"
 description = "Python SDK for UP42, the geospatial marketplace and developer platform."
-authors = ["UP42 GmbH"]
+authors = ["UP42 GmbH <support@up42.com>"]
 license = "https://github.com/up42/up42-py/blob/master/LICENSE"
```
javidq, line 6: We need to change pyproject, otherwise it doesn't have a static license which is problematic since we can change the licensing affecting old versions

javidq: Setup.py just used MIT

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "up42-py"
 version = "0.33.1"
```
javidq, line 3: do we still need up42/_version.txt?

setup.py
```diff
@@ -1,52 +0,0 @@
...
-            ],
-        ),
-    ],
-    include_package_data=True,
-    zip_safe=False,
-    install_requires=parent_dir.joinpath("requirements.txt").read_text().splitlines(),
-    extras_require={
-        "viz": parent_dir.joinpath("requirements-viz.txt").read_text().splitlines(),
-    },
-    classifiers=[
```
javidq, line 44: Missing in pyproject

javidq (changes requested):

seedlit:
> When you publish, can you make sure that reading the sample geometries is still working?
> 
> I mean, the below configuration in `setup.py` made sure that the `aoi_berlin.geojson` and `aoi_washington.geojson` were copied in the published version. We also had to add this in conda. I am not sure if by default these are added in poetry. (_version.txt should be removed)
> 
> ```
> package_data={"": ["_version.txt", "data/aoi_berlin.geojson", "data/aoi_washington.geojson"]}
> ```

I verified this.
It's working. Since the `data` folder is inside `up42` folder, these files are added in the package.
Just in case you want to play around, you can check test it out [here](https://test.pypi.org/project/up42-py-test/) :)

pyproject.toml
seedlit: @Javid do you have a proposal for solution?
I do not want to add the entire [license string](https://github.com/up42/up42-py/blob/master/LICENSE) here.

Makefile
seedlit: > Do we actually need this as a separate step?

Not in the scope of this PR

> serve - better to switch to poetry to make sure that it runs with the specified dependencies

> gh-pages - poetry

Couldn't find the equivalent poetry command. Can you please re-direct me to relevant documentation.

pyproject.toml
seedlit: removed

javidq (changes requested):

up42/utils.py
```diff
@@ -404,3 +405,14 @@ def replace_page_query(url: str, new_page: int) -> str:
     )
 
     return modified_url
+
+
+def get_up42_py_version():
```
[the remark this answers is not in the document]
seedlit: Nice!

pyproject.toml
```diff
@@ -2,12 +2,19 @@
...
 description = "Python SDK for UP42, the geospatial marketplace and developer platform."
-authors = ["UP42 GmbH"]
+authors = ["UP42 GmbH <support@up42.com>"]
 license = "https://github.com/up42/up42-py/blob/master/LICENSE"
 readme = "README.md"
 packages = [
     { include = "up42", from = "." },
 ]
+classifiers=[
```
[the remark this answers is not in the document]
seedlit: coming from pre-commit checks

Makefile
```diff
@@ -1,32 +1,20 @@
...
 	-rm -r .pytest_cache
-	python -m pytest --cov=up42/ --runlive --durations=5
+	poetry run python -m pytest --cov=up42/ --runlive --durations=5
 
 e2e:
 	rm -rf project_20abe*/
-	python $(SRC)/tests/test_e2e_30sec.py
-	python $(SRC)/tests/test_e2e_catalog.py
+	poetry run python $(SRC)/tests/test_e2e_30sec.py
+	poetry run python $(SRC)/tests/test_e2e_catalog.py
```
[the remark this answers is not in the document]
seedlit: It can.
But it makes the command cryptic.
I will skip this one.

.pre-commit-config.yaml
```diff
@@ -40,3 +40,4 @@ repos:
           - "types-requests"
           - "sqlalchemy2-stubs"
           - "types-attrs"
+          - "types-toml"
```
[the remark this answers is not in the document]
seedlit: now not needed anymore since we would not be using toml.
But it was needed because of [failing workflow.](https://app.circleci.com/pipelines/github/up42/up42-py/3671/workflows/f6516398-49fd-4ff3-b683-39371e4a2f2a/jobs/4357)

seedlit:
> https://python-poetry.org/docs/pyproject/#license - can be just "MIT"

I prefer to redirect to proper license file rather than just adding a string "MIT".
I will skip this one.

tests/test_utils.py
```diff
@@ -288,3 +290,9 @@ def test_autocomplete_order_parameters():
     assert order_parameters["params"]["existing_param1"] is not None
     assert order_parameters["params"]["geometry"] is None
     assert order_parameters["params"]["acquisitionMode"] is None
+
+
+@patch("importlib.metadata.version", return_value="0.33.0")
```
javidq, line 295: NIT: this can be confusing and feel like it needs to be updated. I'd still recommend to set it to 0.0.0 or "some-version"

.circleci/config.yml
```diff
@@ -87,18 +87,12 @@ jobs:
           command: poetry run python -m pytest --runlive --durations=5
```
javidq, line 87: Need to update running non-live tests as well

tests/test_auth.py
```diff
@@ -4,6 +4,8 @@
 import pytest
 import requests
 
+from up42.utils import get_up42_py_version
```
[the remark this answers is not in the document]
javidq: `get_up42_py_version` depends on the package to be installed in Venv, otherwise it will fail with PackageNotFound. So we need to mock it.
That's why I'd add to conftest
```python
@pytest.fixture(autouse=True)
def mock_up42_py_version():
  with patch("importlib.metadata.version", return_value="test-version"):
    yield
```
And no changes are needed here, sorry for confusion

javidq (changes requested):

javidq (approved):

PR #508, opened 2023-12-05
Extract endpoint functionality to a separate module

javidq:
`Auth::_endpoint` and its usage is extracted to the dedicated module `host.py`, while keeping environment switch in Auth class to avoid breaking sdk tests. The change is motivated by simplification of http related functionality to improve testability of the design. The functionality and consumer API hasn't been changed in the refactoring process.

Items:
* [x] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/auth.py
```diff
@@ -197,15 +193,23 @@ def _get_token_project_based(self):
...
         self.token = token_response["data"]["accessToken"]
 
+    @property
+    def env(self):
+        return host.DOMAIN
+
+    @env.setter
+    def env(self, value: str):
+        host.DOMAIN = value
+
```
javidq, line 209: This change is not needed for end users, but it's needed for sdk tests

javidq: Will be dropped when sdk tests are adapted

seedlit:
@javidq since we are hardcoding the domain in `host.py`, would it still work for staging environment?

javidq:
> @javidq since we are hardcoding the domain in `host.py`, would it still work for staging environment?

Yes, since the keep the same interface via env parameter in auth construction. After this PR is merged I will create a PR on SDK tests to manipulate host.env directly and drop it from auth completely

seedlit (approved):
looked over the surface.
LGTM!

PR #510, opened 2023-12-08
AT-3182 Migrating estimate and create orders

andher1802:
Migrating create and estimate orders to V2

[AT-3182](https://up42.atlassian.net/browse/AT-3182?focusedCommentId=36308&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-36308)

Items:
* [X] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

[AT-3182]: https://up42.atlassian.net/browse/AT-3182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/test_order.py
```diff
@@ -215,17 +232,61 @@ def test_track_status_fail(order_mock, status, requests_mock):
...
-        f"{auth_mock._endpoint()}/workspaces/{auth_mock.workspace_id}/orders/estimate"
-    )
-    requests_mock.post(url=url_order_estimation, json={"data": {"credits": 100}})
+    url_order_estimation = f"{auth_mock._endpoint()}/v2/orders/estimate"
+    expected_payload = {
+        "summary": {"totalCredits": 38, "totalSize": 0.1, "unit": "SQ_KM"},
+        "results": [{"index": 0, "credits": 38, "unit": "SQ_KM", "size": 0.1}],
+        "errors": [],
+    }
+    expected_output = {
```
janchrizz, line 241: nitpick: it feels like this exact dict is repeated 2-3 times, maybe makes sense to make it into a variable?

up42/catalog.py
```diff
@@ -414,8 +431,7 @@ def construct_order_parameters(
         # Handled on API level, don't manipulate in SDK, providers might accept geometries in the future.
         if aoi is not None:
             aoi = any_vector_to_fc(vector=aoi)
-            aoi = fc_to_query_geometry(fc=aoi, geometry_operation="intersects")
-            order_parameters["params"]["aoi"] = aoi  # type: ignore
+            order_parameters["featureCollection"] = aoi  # type: ignore
```
janchrizz, line 434: are you sure about this?
CMIIW but it seems the aoi is inside something like `["featureCollection"]["features"][0]["geometry"] `if looking at https://docs.up42.com/developers/api#tag/Orders/operation/createV2OrderUnderWorkspace

andher1802: To my understanding they are now asking for a GeoJSON in the featureCollection element. The aoi should be a geoJSON object and the function any_vector_to_fc converts any type of supported Geometry to featureCollection (fc). If you see now we are not getting the aoi from the feature collection, as it is not used anymore in the endpoint. 

Do you think I missed something?

tests/test_order.py
andher1802: Good question. The idea is to make tests as independent as possible. If we want in the future to change tests I would not like to need to affect the other ones. In this case, Do you think that sharing variables is recommended bad among tests?

up42/catalog.py
andher1802: I will rename this variable.

janchrizz: Ah ok, i want aware of what `fc_to_query_geometry` does. Nvm then, thank you

dustgalactic:
@andher1802 I have a couple of questions/comments before I can support this release from the doc side:

What needs to be changed in the documentation, please confirm:
1. The return data type of [catalogBase.place_order()](https://github.com/up42/up42-py/pull/510/files#diff-7d6090c79f0c2938180b7ebe89ac5015d39aef12cbb9ef0faa047958ec441d12R134) should be changed from `Order` to `List[Order]`.
2. A new parameter should be added to [catalog.construct_order_parameters()](https://github.com/up42/up42-py/pull/510/files#diff-7d6090c79f0c2938180b7ebe89ac5015d39aef12cbb9ef0faa047958ec441d12R380): `extra_params: Optional[dict] = {}`.
3. Following our discussion that the static method [order.estimate()](https://github.com/up42/up42-py/pull/510/files#diff-cd6be07752d0ee55f75a7a8eb11de35ee0a889d69277b06362fb83c19d99c9a9R103) shoudn't be exposed to users and is used inside `catalog.estimate_order()` (like `order.place()` is used inside `catalogBase.place_order()`), then the return data type of `catalog.estimate_order()` should be changed from `int` to `dict`.

Questions:
* We’re not renaming CatalogBase’s [place_order()](https://github.com/up42/up42-py/pull/510/files#diff-7d6090c79f0c2938180b7ebe89ac5015d39aef12cbb9ef0faa047958ec441d12R128) to `create_order()` depsite the inconsistency in the terminology, correct? Re: [discussion](https://up42.atlassian.net/browse/AT-3182?focusedCommentId=34313).
* The current public function that estimates order is `catalog.estimate_order()` — because before the release only catalog orders could be estimated. However, the new [endpoint](https://docs.up42.com/developers/api#tag/Orders/operation/estimateV2Order) allows users to estimate any type of order. It means, that the function should be transferred to the `CatalogBase` class or another class that determines functionalities of both order types.
* Why was [this line](https://github.com/up42/up42-py/pull/510/files#diff-cd6be07752d0ee55f75a7a8eb11de35ee0a889d69277b06362fb83c19d99c9a9R171) updated? `if self.info["type"] == "TASKING" else ""`. [Sub-statuses](https://docs.up42.com/developers/api-tasking/tasking-monitor#order-sub-statuses) are still tasking-only ([catalog statuses](https://docs.up42.com/developers/api-catalog/catalog-monitor)).

tests/test_order.py
janchrizz: Hi yes in general, tests should be independent of each other. But in this case, this 4 occurrences are in the forms of the same function "estimate_order" tests, i think the `expected_output` variable value wont change (if this expected response change in 1, and it doesnt change in the other 3, i think there is something wrong)
But it's just a suggestion.

andher1802:
Hi Daria, 

Thanks for your comments I was not aware of the change of scope of the new order endpoints. You are right, I will adjust the new endpoints accordingly. 

About your questions. I confirm: 
1. 2. and 3. are correct.

- I will rename it to `create_orders`
- I will move to Catalog Base and create a test for estimating tasking orders. 
- AFAIK `sub statuses` apply only to tasking orders, that is what I got this info from the API documentation

> @andher1802 I have a couple of questions/comments before I can support this release from the doc side:
> 
> What needs to be changed in the documentation, please confirm:
> 
> 1. The return data type of [catalogBase.place_order()](https://github.com/up42/up42-py/pull/510/files#diff-7d6090c79f0c2938180b7ebe89ac5015d39aef12cbb9ef0faa047958ec441d12R134) should be changed from `Order` to `List[Order]`.
> 2. A new parameter should be added to [catalog.construct_order_parameters()](https://github.com/up42/up42-py/pull/510/files#diff-7d6090c79f0c2938180b7ebe89ac5015d39aef12cbb9ef0faa047958ec441d12R380): `extra_params: Optional[dict] = {}`.
> 3. Following our discussion that the static method [order.estimate()](https://github.com/up42/up42-py/pull/510/files#diff-cd6be07752d0ee55f75a7a8eb11de35ee0a889d69277b06362fb83c19d99c9a9R103) shoudn't be exposed to users and is used inside `catalog.estimate_order()` (like `order.place()` is used inside `catalogBase.place_order()`), then the return data type of `catalog.estimate_order()` should be changed from `int` to `dict`.
> 
> Questions:
> 
> * We’re not renaming CatalogBase’s [place_order()](https://github.com/up42/up42-py/pull/510/files#diff-7d6090c79f0c2938180b7ebe89ac5015d39aef12cbb9ef0faa047958ec441d12R128) to `create_order()` depsite the inconsistency in the terminology, correct? Re: [discussion](https://up42.atlassian.net/browse/AT-3182?focusedCommentId=34313).
> * The current public function that estimates order is `catalog.estimate_order()` — because before the release only catalog orders could be estimated. However, the new [endpoint](https://docs.up42.com/developers/api#tag/Orders/operation/estimateV2Order) allows users to estimate any type of order. It means, that the function should be transferred to the `CatalogBase` class or another class that determines functionalities of both order types.
> * Why was [this line](https://github.com/up42/up42-py/pull/510/files#diff-cd6be07752d0ee55f75a7a8eb11de35ee0a889d69277b06362fb83c19d99c9a9R171) updated? `if self.info["type"] == "TASKING" else ""`. [Sub-statuses](https://docs.up42.com/developers/api-tasking/tasking-monitor#order-sub-statuses) are still tasking-only ([catalog statuses](https://docs.up42.com/developers/api-catalog/catalog-monitor)).

dustgalactic:
@andher1802 Thanks for the clarification!

> I will rename it to create_orders

I'm not sure about the plural form in the name of the function. If it's a direct counterpart of the [Create an order](https://docs.up42.com/developers/api#tag/Orders/operation/createV2OrderUnderWorkspace) endpoint, then I suggest `create_order()`. But if several orders can be created, then sure, `create_orders()` is better in this case.

> AFAIK sub statuses apply only to tasking orders, that is what I got this info from the API documentation

Yes, you're correct. It's just that I thought that your changes now include sub-statuses for catalog orders as well, which would be incorrect.

> I will rename it to create_orders
I will move to Catalog Base and create a test for estimating tasking orders.

Please tag me here or on Slack after you've finished with the changes, I'll update all the necessary docs in the following commit (it won't take long).

PR #528, opened 2024-01-20
[AT-3197] Adding deprecation message VIZ features

andher1802:
Adding deprecation warnings to the visualization features in Job, JobCollection.

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


dustgalactic:
@andher1802 Should I add a deprecation banner to instances of these functions throughout the reference + remove the https://sdk.up42.com/visualizations/ page?

up42/viztools.py
```diff
@@ -137,6 +138,12 @@ def plot_results(
                  e.g. matplotlib cmap etc.
         """
         warnings.filterwarnings("ignore", category=rasterio.errors.NotGeoreferencedWarning)
+        warn(
+            "Visualization methods are getting deprecated. The current feature will be discontinued "
+            "after March 31, 2024, and will be replaced by new processing functionalities.",
```
seedlit, line 143: If I am not wrong, we aren't providing any alternative for visualization in new processing functionalities right?
In this case, consider removing "and will be replaced by new processing functionalities."

seedlit: same for others below.

seedlit (commented):
left a comment :)

up42/viztools.py
andher1802: I will add your suggestion thanks. I think adding new functionalities can be redundant.

But please be aware that this deprecation message will be exposed when the user runs this function from Job, JobCollection, and JobTask classes. I was referring to these classes when I added the new functionalities. Just to make sure we are all aware of the context.

seedlit: I see.
Maybe we can control that by updating the `stacklevel=2` param? 🤔

dustgalactic:
@andher1802 I've removed the page about visualization and mentions of the functionality throughout the documentation. As for functions, I've added a "This function is deprecated." banner. IMHO, this is different from the discontinuation announcement we have + the Catalog class also uses some of these functions.

up42/viztools.py
```diff
@@ -137,6 +138,12 @@ def plot_results(
                  e.g. matplotlib cmap etc.
         """
         warnings.filterwarnings("ignore", category=rasterio.errors.NotGeoreferencedWarning)
+        warn(
+            "Visualization methods are getting deprecated. The current feature will be discontinued "
```
dustgalactic, line 142: Visualization methods are getting deprecated -> Visualization methods are deprecated

seedlit (approved):
LGTM!
Just one question: stacklevel=2 means the error will be surfaced in the parent function right?

PR #536, opened 2024-02-12
[AT-4372] Configure mypy: asset.py and catalog.py

seedlit:
Jira ticket [here](https://up42.atlassian.net/browse/AT-4372).
NOTE: most of the changes are coming from poetry files.
Scope: 
 - temporarily use mypy.ini file for configuration
 - fix asset.py (ticket [here](https://up42.atlassian.net/browse/AT-4375))
 - fix catalog.py (ticket [here](https://up42.atlassian.net/browse/AT-4378))

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


.pre-commit-config.yaml
```diff
@@ -28,15 +28,19 @@ repos:
...
-    rev: "v1.1.1" # Use the sha / tag you want to point at
+  - repo: local
     hooks:
       - id: mypy
-        args: [--no-strict-optional, --ignore-missing-imports]
+        files: \.py$
+        name: mypy
+        entry: mypy
+        language: python
+        args: ["--config-file=mypy.ini"]
```
javidq, line 38: Not needed

.pre-commit-config.yaml
```diff
@@ -28,15 +28,19 @@ repos:
...
         additional_dependencies:
           - "types-PyYAML"
           - "types-requests"
           - "sqlalchemy2-stubs"
           - "types-attrs"
           - "types-python-dateutil"
           - "types-mock"
+          - "types-tqdm"
```
javidq, line 46: These 3 lines are not needed, since they are in dev dependencies, 4 lines above should move to dev dependencies as well

mypy.ini
```diff
@@ -0,0 +1,107 @@
...
+[mypy-up42.webhooks]
+ignore_errors = True
+[mypy-up42.workflow]
+ignore_errors = True
+[mypy-tests.conftest]
+ignore_errors = True
+[mypy-tests.context]
+ignore_errors = True
+[mypy-tests.test_asset]
+ignore_errors = True
```
javidq, line 51: We must enable test_asset.py together with asset.py and the same applies to every module

mypy.ini
```diff
@@ -0,0 +1,107 @@
...
+[mypy-tests.test_auth]
+ignore_errors = True
+[mypy-tests.test_catalog]
+ignore_errors = True
```
javidq, line 55: The same as above. I'd recommend to set catalog.py ignore_errors=True and focus on asset .py and test_asset.py

mypy.ini
javidq: `In the child issues below each “Typing hints“ task must cover both implementation code file and the corresponding test.` from task description

mypy.ini
```diff
@@ -0,0 +1,107 @@
+[mypy]
+check_untyped_defs = True
+strict_optional = True
+[mypy-docs.macros]
+ignore_errors = True
+[mypy-up42.asset]
+ignore_errors = False
```
javidq, line 7: We don't need to add a module with False, it's default. On each step we're supposed to remove a module and the corresponding test module from this file

up42/catalog.py
```diff
@@ -6,14 +6,14 @@
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
-from geojson import Feature, FeatureCollection
-from geopandas import GeoDataFrame
-from shapely.geometry import Polygon
+from geojson import Feature, FeatureCollection  # type: ignore
+from geopandas import GeoDataFrame  # type: ignore
+from shapely.geometry import Polygon  # type: ignore
 from tqdm import tqdm
```
javidq, line 13: I don't understand the reason to introduce these changes since when I run mypy check on the master version it raises only 3 errors related to implicit optional

javidq (changes requested):

mypy.ini
seedlit: I see.
For this PR, let's go ahead with both: asset and catalog and their corresponding tests.

up42/catalog.py
seedlit: If I do not add these, I see these 3 errors (similar to what you are getting I guess)?
Do you have any alternative solution?

image

javidq: I didn't mean imported module, I meant the changes in the code

seedlit: I could not reproduce.
Reverted the un-necessary changes.

PR #538, opened 2024-02-14
AT-4219 Fixing tasking order translate bug

andher1802:
Fixing the OrderV2 migration to allow tasking orders.

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/tests/fixtures/fixtures_order.py
+++ b/tests/fixtures/fixtures_order.py
@@ -71,7 +71,7 @@ def order_live(auth_live):
 
 
 @pytest.fixture
-def order_parameters():
+def catalog_order_parameters():
     return {
         "dataProduct": "4f1b2f62-98df-4c74-81f4-5dce45deee99",
         "params": {
@@ -96,7 +96,7 @@ def order_parameters():
 
 
 @pytest.fixture
-def order_parameters_tasking():
+def tasking_order_parameters():
     return {
         "dataProduct": "47dadb27-9532-4552-93a5-48f70a83eaef",
         "params": {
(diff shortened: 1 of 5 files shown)


tests/test_order.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 
 from up42.asset import Asset
-from up42.order import Order
+from up42.order import Order, _translate_construct_parameters
```
javidq, line 6: We don't test the hidden functions. Please parameterize the tests of order placement and estimation for the case of tasking

javidq (changes requested):

tests/fixtures/fixtures_order.py
```diff
@@ -93,3 +93,41 @@ def order_parameters():
...
+            "spectralProcessing": None,
+            "pixelCoding": None,
+            "radiometricProcessing": None,
+            "deliveredAs": None,
+        },
+    }
+
+
+@pytest.fixture(params=["catalog", "tasking"])
+def order_parameters_tasking_catalog(request, catalog_order_parameters, tasking_order_parameters):
```
javidq, line 131: Just order_parameters, pytest will display params anyway

tests/test_catalog.py
```diff
@@ -457,25 +457,32 @@ def test_estimate_order_from_catalog(order_parameters, requests_mock, auth_mock)
...
-    estimation = catalog_instance.estimate_order(order_parameters)
+    estimation = catalog_instance.estimate_order(catalog_order_parameters)
     assert isinstance(estimation, int)
     assert estimation == 100
 
 
-def test_order_from_catalog(order_parameters, order_mock, catalog_mock, requests_mock):
+def test_order_from_catalog(
+    catalog_order_parameters,
+    tasking_order_parameters,
```
javidq, line 467: Why do we need these?

javidq (changes requested):

tests/test_catalog.py
andher1802: You are right, I have added those to test separately and forgot to remove them at the end.

PR #541, opened 2024-02-19
Configure mypy 2

seedlit:
Fixes mypy configuration for tools.py, utils.py, tasking.py and their corresponding tests.

Jira tickets here:
 - https://up42.atlassian.net/browse/AT-4389
 - https://up42.atlassian.net/browse/AT-4390
 - https://up42.atlassian.net/browse/AT-4391

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/utils.py
```diff
@@ -388,7 +387,7 @@ def replace_page_query(url: str, new_page: int) -> str:
     """
     parsed_url = urlparse(url)
     query_params = parse_qs(parsed_url.query)
-    query_params["page"] = str(new_page)  # type: ignore
+    query_params["page"] = [str(new_page)]
```
javidq, line 390: Why do the change type here?

javidq (changes requested):

up42/utils.py
seedlit: Otherwise mypy throws an error:
 `error: Incompatible types in assignment (expression has type "str", target has type "list[str]")  [assignment]`
 
 If, I am not mistaken, this requirement is coming from the `parse_qs` function.
 
image

javidq (approved):

PR #545, opened 2024-02-19
[AT-4372] Configure mypy 4

seedlit:
Scope: Fixes (mypy) - `macros.py`, `auth.py`, storage.py`, `estimate.py` and their corresponding tests.

Note: I will create another PR for final cleanups in remaining scripts.

Items:
* [ ] Ran test & live-tests
* [ ] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog


up42/auth.py
```diff
@@ -64,7 +64,11 @@ class retry_if_429_rate_limit(retry_if_exception):
     def __init__(self):
         def is_http_429_error(exception):
             if exception.response is not None:
-                return isinstance(exception, requests.exceptions.HTTPError) and exception.response.status_code == 429
+                return (
```
javidq, line 67: Seems like master merge is needed

up42/storage.py
```diff
@@ -220,12 +220,12 @@ def get_orders(
             "status": set(statuses) & allowed_statuses if statuses else None,
         }
         params = {k: v for k, v in params.items() if v is not None}
-
-        if statuses is not None and len(statuses) > len(params["status"]):
```
javidq, line 224: This whole block is not needed, let's just drop it

javidq (changes requested):

tests/fixtures/fixtures_order.py
```diff
@@ -67,7 +67,9 @@ def order_mock(auth_mock, requests_mock):
 
 @pytest.fixture()
 def order_live(auth_live):
-    return Order(auth=auth_live, order_id=os.getenv("TEST_UP42_ORDER_ID"))
+    order_id = os.getenv("TEST_UP42_ORDER_ID")
+    assert isinstance(order_id, str)
```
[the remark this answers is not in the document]
seedlit: if I remove the assertion, mypy complains with
`ests/fixtures/fixtures_order.py:72: error: Argument "order_id" to "Order" has incompatible type "str | None"; expected "str"  [arg-type]`

in line `return Order(auth=auth_live, order_id=order_id)`

Happy to receive any suggestions.

up42/storage.py
seedlit: do we not want to add logs when statuses is not included in allowed_stasuses?

javidq: I don't see any reason for that - if a user uses an IDE, then type hints will show incorrect statuses, if not, then they need to rely on documentation. Also this does not create any error

docs/macros.py
```diff
@@ -22,12 +23,14 @@ def define_env(env):
     This is the hook for defining variables, macros and filters.
     """
 
-    def indent(input_string: str):
-        return input_string.replace("\n", "\n\t")
+    def indent(input_string: Optional[str]) -> Optional[str]:
+        if input_string:
```
javidq, line 27: Can be simplified to `return input_string and input_string.replace("\n", "\n\t")`

tests/fixtures/fixtures_order.py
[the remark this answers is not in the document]
javidq: Strange, I run mypy on `return Order(auth=auth_live,order_id=os.environ["TEST_UP42_ORDER_ID"])` without any failures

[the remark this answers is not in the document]
seedlit: indeed.
I was probably doing some stupid mistake.
Updated.

docs/macros.py
seedlit: you probably meant `or`, otherwise it returns None.
I will update `return input_string or input_string.replace("\n", "\n\t")`

tests/fixtures/fixtures_auth.py
```diff
@@ -56,6 +56,12 @@ def auth_mock(request, auth_project_mock, auth_account_mock):
     return mocks[request.param]
 
 
+@pytest.fixture()
```
javidq, line 59: Merge artefact, not needed

tests/conftest.py
```diff
@@ -13,6 +13,7 @@
     password_test_live,
     project_api_key_live,
     project_id_live,
+    setup_auth_mock,
```
javidq, line 16: Not needed

javidq (changes requested):

javidq (approved):

PR #550, opened 2024-03-06
AT-4866 Resilient and authenticated session

javidq:
Porting resilient and authenticated session

Items:
* [x] Ran test & live-tests
* [x] Removed credentials

For release:
* [x] Bumped version
* [x] Added changelog

### DIFF
--- a/tests/http/test_session.py
+++ b/tests/http/test_session.py
@@ -1,0 +1,62 @@
+import random
+from unittest import mock
+
+import pytest
+import requests
+from requests import adapters
+from requests_mock import Mocker
+
+from up42.http import session as up42_session
+
+some_url = "https://something.com"
+
+methods_and_calls = [
+    ("GET", requests.Session.get),
+    ("POST", requests.Session.post),
+    ("PUT", requests.Session.put),
+    ("PATCH", requests.Session.patch),
+    ("DELETE", requests.Session.delete),
+    ("OPTIONS", requests.Session.options),
+    ("HEAD", requests.Session.head),
+]
+
+authorization_value = "Bearer some-token"
+
+
+def set_token(request: requests.Request):
+    request.headers["Authorization"] = authorization_value
+    return request
+
+
+create_auth = mock.MagicMock(return_value=mock.MagicMock(side_effect=set_token))
+create_adapter = mock.MagicMock(return_value=adapters.HTTPAdapter())
+
+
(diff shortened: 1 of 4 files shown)


tests/http/test_session.py
```diff
@@ -0,0 +1,62 @@
...
+
+authorization_value = "Bearer some-token"
+
+
+def set_token(request: requests.Request):
+    request.headers["Authorization"] = authorization_value
+    return request
+
+
+create_auth = mock.MagicMock(return_value=mock.MagicMock(side_effect=set_token))
```
seedlit, line 31: out of curiosity - why not define these inside the `create_session` fixture?
(since they are only used inside the fixture)

tests/http/test_session.py
```diff
@@ -0,0 +1,62 @@
...
+    ("GET", requests.Session.get),
+    ("POST", requests.Session.post),
+    ("PUT", requests.Session.put),
+    ("PATCH", requests.Session.patch),
+    ("DELETE", requests.Session.delete),
+    ("OPTIONS", requests.Session.options),
+    ("HEAD", requests.Session.head),
+]
+
+authorization_value = "Bearer some-token"
```
seedlit, line 23: since we are defining it outside the test, and using it at multiple places, this looks more like a global constant.
Consider renaming the variable to `AUTHORIZATION_VALUE`

seedlit: same comment for `some_url`

up42/http/session.py
```diff
@@ -0,0 +1,29 @@
...
+
+@lru_cache
+def create(
+    create_adapter=http_adapter.create,
+    create_auth=oauth.ProjectAuth,
+) -> requests.Session:
+    session = StatusValidatingSession()
+    adapter = create_adapter()
+    for schema in SCHEMAS:
+        session.mount(schema + "://", adapter)
```
javidq, line 27: This part is not covered by tests, since it requires an integration test on network level, which is not supported by requests_mock

seedlit (commented):
left some comments.
Generic remark: since we are adding new functions, consider adding doc-strings so we don't create tech debt over time.

tests/http/test_session.py
javidq: No strong opinion here, moved it inside

seedlit (approved):

dustgalactic (approved):

PR #556, opened 2024-03-07
AT-4080 Enable pylint for auth and conf files

javidq:
Drop linting exclusion for auth and conf files

Items:
* [x] Ran test & live-tests
* [x] Removed credentials

For release:
* [x] Bumped version
* [x] Added changelog

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -1,20 +1,20 @@
-import json
-from pathlib import Path
+import pathlib
+from typing import cast
 
 import pytest
 import requests
+from requests_mock import Mocker
 
-from up42 import host
+from up42 import host, utils
 from up42.auth import Auth
-from up42.utils import get_up42_py_version
 
-from .fixtures.fixtures_globals import PROJECT_APIKEY, PROJECT_ID, TOKEN, WORKSPACE_ID
+from .fixtures import fixtures_globals as constants
 
 
 def test_auth_kwargs():
     auth = Auth(
-        project_id=PROJECT_ID,
-        project_api_key=PROJECT_APIKEY,
+        project_id=constants.PROJECT_ID,
+        project_api_key=constants.PROJECT_APIKEY,
         env="abc",
         authenticate=False,
         retry=False,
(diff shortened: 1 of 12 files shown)


tests/test_auth.py
```diff
@@ -26,81 +26,54 @@ def test_auth_kwargs():
...
 
 
-def test_get_token(auth_mock):
-    auth_mock._get_token()
-    assert auth_mock.token == TOKEN
+def test_get_workspace(auth_mock: up42_auth.Auth):
+    assert auth_mock.workspace_id == constants.WORKSPACE_ID
 
 
-@pytest.mark.live
```
javidq, line 54: @janchrizz please confirm that it's safe to drop these tests

janchrizz: Yes we can drop these auth tests

janchrizz (approved):
my approval doesnt count but here it is :D

up42/auth.py
```diff
@@ -2,33 +2,24 @@
...
-    retry_if_exception,
-    retry_if_exception_type,
-    stop_after_attempt,
-    wait_fixed,
-    wait_random_exponential,
-)
+import requests_oauthlib as roauthlib  # type:ignore
+import tenacity as tnc
+from oauthlib import oauth2
+from requests import auth as rauth
```
seedlit, line 14: why not just use `requests.auth`?

up42/auth.py
```diff
@@ -2,33 +2,24 @@
...
 """
 import json
-from pathlib import Path
+import pathlib
+import warnings
 from typing import Dict, Optional, Union, cast
-from warnings import warn
 
 import requests
 import requests.exceptions
```
seedlit, line 10: since we are already importing requests, maybe we can remove this import?

up42/auth.py
```diff
@@ -70,16 +61,13 @@ def __init__(
...
-            "password",
-        ]
         credentials_dict = {
-            key: value for key, value in local_vars.items() if key in credentials_filter and value is not None
+            "project_id": project_id,
+            "project_api_key": project_api_key,
+            "username": username,
+            "password": password,
         }
+        credentials_dict = {key: value for key, value in credentials_dict.items() if value}
```
seedlit, line 70: nitpick: this will probably never be the case, but since we aren't validating, in theory it is possible to set a bool value to any of the key in `credentials_dict`

If any value is set to `False`, `credentials_dict = {key: value for key, value in credentials_dict.items() if value}` --> this logic would fail.

can we bring back `is not None`?

seedlit (commented):
Nice!
left some comments.

up42/auth.py
javidq: That doesn't make sense for the case of credentials. If some users want authentication to fail, completely ignoring the documentation, let them fail

seedlit (approved):

PR #560, opened 2024-03-12
AT4927 [Hotfix] Asset update metadata

andher1802:

Items:
* [x] Ran test & live-tests
* [x] Removed credentials

For release:
* [X] Bumped version
* [X] Added changelog


up42/asset.py
```diff
@@ -103,14 +103,24 @@ def update_metadata(
         Update the metadata of the asset.
 
         Args:
-            title: The title string to be assigned to the asset.
-            tags: A list of tag strings to be assigned to the asset.
+            title: The title string to be assigned to the asset. No value will keep the existing title.
```
seedlit, line 106: Maybe we can improve this part.
I am not sure what do we mean by `"No value will keep the existing title."`
Same for the next line.
@dustgalactic any suggestions?

up42/asset.py
```diff
@@ -103,14 +103,24 @@ def update_metadata(
...
+            tags: A list of tag strings to be assigned to the asset. No value will keep the existing tags.
 
         Returns:
             The updated asset metadata information
         """
         url = endpoint(f"/v2/assets/{self.asset_id}/metadata")
-        body_update = {"title": title, "tags": tags, **kwargs}
+        if kwargs:
+            logger.info(
+                f"{','.join([key for key in kwargs.keys()])} values are not allowed to update in asset metadata."
```
seedlit, line 115: if we are not allowing any key from kwargs, should we not just drop the `kwargs` argument altogether?

tests/test_asset.py
```diff
@@ -80,6 +84,18 @@ def test_asset_update_metadata(asset_mock):
     assert updated_info["tags"] == ["othertag1", "othertag2"]
 
 
+def test_asset_update_metadata_should_return_same_with_no_values(asset_mock):
```
seedlit, line 87: is it not similar to `test_asset_update_metadata`?

tests/test_asset.py
```diff
@@ -80,6 +84,18 @@ def test_asset_update_metadata(asset_mock):
...
+    updated_info = asset_mock.update_metadata(title="some_other_title", tags=["othertag1", "othertag2"])
+    assert updated_info["title"] == asset_mock.info["title"]
+    assert updated_info["tags"] == asset_mock.info["tags"]
+
+
+def test_asset_update_metadata_should_ignore_kwargs(asset_mock, caplog):
```
seedlit, line 93: nice with the caplog!
didn't knew about this :)

seedlit (commented):
left some comments :)

up42/asset.py
andher1802: no change in method signature, otherwise we break user code.

tests/test_asset.py
andher1802: nop, one tests if the data is updated, the other test the other case (that was the source of the error), if not provided then no update return the same.

up42/asset.py
andher1802: It means that if the user does not provide a title or tag it will no update the field. In the previous case (that it was the actual bug we are trying to solve) it removed title/tags.

CHANGELOG.md
```diff
@@ -30,6 +30,13 @@ You can check your current version with the following command:
 
 For more information, see [UP42 Python package description](https://pypi.org/project/up42-py/).
 
+## 0.37.0a5
```
dustgalactic, line 33: This will be moved to another file as per this PR: https://github.com/up42/up42-py/pull/554

tests/test_asset.py
andher1802: You are right, I stupidly did not remove the inputs that were what I wanted to test. Thanks.

up42/asset.py
dustgalactic: Changes to these files won't update the reference: https://sdk.up42.com/reference/asset-reference/#update_metadata

What we have on the reference right now:

* title: An editable asset title.
* tags: An editable list of tags to categorize the asset.

These arguments are optional. An "optionality" in an argument behavior means that nothing will be changed. My question is, why do you think we need to update these texts?

up42/asset.py
```diff
@@ -103,14 +103,24 @@ def update_metadata(
...
-            tags: A list of tag strings to be assigned to the asset.
+            title: The title string to be assigned to the asset. No value will keep the existing title.
+            tags: A list of tag strings to be assigned to the asset. No value will keep the existing tags.
 
         Returns:
             The updated asset metadata information
         """
         url = endpoint(f"/v2/assets/{self.asset_id}/metadata")
-        body_update = {"title": title, "tags": tags, **kwargs}
+        if kwargs:
```
janchrizz, line 113: if say a user intends to actually remove the title and tags, can he still do it by passing an empty string or null?

up42/asset.py
andher1802: Because there is an error that is fixed as we can't change the signature. Optional and = None means that the value is not optional anymore. It means that when the user does not set the `title` the value is set to None by default. The correct behavior is that the `title` is not set to any value.

The unwanted behavior that we have right now is that when the user does not set `title` and `title` is set to None, then the update metadata removes the asset title. The behavior we want is that when the user does not set `tilte` value and the `title` is set to None, then the update metadata ignores this field and keep the title as it is.

up42/asset.py
andher1802: Nop, it will be ignored. That is the tradeoff of the solution. We should add an argument to the signature to remove titles. That is the proposal to solve the issue. Maybe there is a better way :)

up42/asset.py
dustgalactic: Ok, will update the reference in this PR.

up42/asset.py
andher1802: Screenshot 2024-03-13 at 14 17 57

javidq: Please just deprecate it, this block is not needed

up42/asset.py
javidq: we need to change the default value here from None to something like [this](https://github.com/Python-Optional/optional.py/blob/main/optional/nothing.py). Please add it to the utils with the corresponding tests. This way we will be able to separate the case when None value is provided versus the case when it's not provided at all

javidq (changes requested):

up42/asset.py
andher1802: I have updated following a simpler approach but with the same results. Please check it and let me know what you think. Also, I have update the tests.

up42/asset.py
```diff
@@ -93,24 +96,30 @@ def stac_items(self) -> pystac.ItemCollection:
...
         Args:
-            title: The title string to be assigned to the asset.
-            tags: A list of tag strings to be assigned to the asset.
+            title: The title string to be assigned to the asset. No value will keep the existing title.
+            tags: A list of tag strings to be assigned to the asset. No value will keep the existing tags.
 
         Returns:
             The updated asset metadata information
         """
+        if kwargs:
```
javidq, line 113: @dustgalactic @kodamaaiko if the usage of kwargs arguments was never documented and doesn't work anyway for normal users, can we just drop them?

tests/test_asset.py
```diff
@@ -1,10 +1,13 @@
...
 import pathlib
+from typing import Dict, List, Optional
 from unittest import mock
 
 import pystac
 import pytest
+import requests_mock as req_mock
 
-from up42 import asset, host
+from up42 import asset, auth, host
```
seedlit, line 10: nitpick: although not necessary, but consider importing `auth as up42_auth` to be consistent with imports in other modules.

seedlit (approved):
left a non-blocking comment.

up42/asset.py
dustgalactic: Makes sense to me.

kodamaaiko: I dont know the full implication, but if it's not documented and users arent using them, I say yes, lets drop.

seedlit (approved):

javidq (approved):

PR #563, opened 2024-03-13
Enable linting for webhooks

javidq:
Enabling linting for webhooks + tests + fixtures

Items:
* [x] Ran test & live-tests
* [x] Removed credentials

For release:
* [ ] Bumped version
* [ ] Added changelog

### DIFF
--- a/tests/test_webhooks.py
+++ b/tests/test_webhooks.py
@@ -1,19 +1,19 @@
 import pytest
 
-from up42.webhooks import Webhook
+from up42 import webhooks
 
-from .fixtures.fixtures_globals import WEBHOOK_ID, WORKSPACE_ID
+from .fixtures import fixtures_globals as constants
 
 
 def test_webhook_initiate(webhook_mock):
-    assert isinstance(webhook_mock, Webhook)
-    assert webhook_mock.webhook_id == WEBHOOK_ID
-    assert webhook_mock.workspace_id == WORKSPACE_ID
+    assert isinstance(webhook_mock, webhooks.Webhook)
+    assert webhook_mock.webhook_id == constants.WEBHOOK_ID
+    assert webhook_mock.workspace_id == constants.WORKSPACE_ID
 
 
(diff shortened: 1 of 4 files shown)


tests/test_webhooks.py
```diff
@@ -1,19 +1,17 @@
...
+    assert isinstance(webhook_mock, webhooks.Webhook)
+    assert webhook_mock.webhook_id == constants.WEBHOOK_ID
+    assert webhook_mock.workspace_id == constants.WORKSPACE_ID
 
 
 def test_webhook_info(webhook_mock):
-    assert webhook_mock.info
-    assert webhook_mock._info["id"] == WEBHOOK_ID
+    info = webhook_mock.info
+    assert info and info["id"] == constants.WEBHOOK_ID
```
seedlit, line 14: will `info and info["id"]` not always return `info["id"]`?

If that's the case, consider updating assertion to
`assert info["id"] == constants.WEBHOOK_ID`

tests/test_webhooks.py
```diff
@@ -24,8 +22,8 @@ def test_webhook_trigger_test_event(webhook_mock):
 
 def test_webhook_update(webhook_mock):
     updated_webhook = webhook_mock.update(name="test_info_webhook")
-    assert isinstance(updated_webhook, Webhook)
-    assert updated_webhook._info["name"] == "test_info_webhook"
+    assert isinstance(updated_webhook, webhooks.Webhook)
+    assert "test_info_webhook" in repr(updated_webhook)
```
seedlit, line 26: nice!

seedlit (commented):
nice!
left a comment.

tests/test_webhooks.py
javidq: info and info["id"] always return info["id"] since it always uses the last value if the first one can be evaluated as True

seedlit (approved):

PR #565, opened 2024-03-13
[AT-4915] Http client

javidq:
New http stack client to provide resilient expiration managing token and requests compliant authentication.

Items:

For release:
* [x] Bumped version
* [x] Added changelog


up42/http/client.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def _merge(left: Optional[config.CredentialsSettings], right: Optional[config.CredentialsSettings]):
+    if all([left, right]):
+        raise MultipleCredentialsSources("Multiple sources of credentials provided")
+    return left or right
+
+
+def create(
+    credential_sources: List[Dict],
```
seedlit, line 23:
```suggestion
    credential_sources: list[dict],
```
since we won't be supporting python 3.8 anymore, we can just use the default dict and list objects natively.

up42/http/client.py
```diff
@@ -0,0 +1,42 @@
...
+    token_url: str,
+    detect_settings=oauth.detect_settings,
```
seedlit, line 25: let's use type hints for all the arguments.

up42/http/client.py
seedlit: same for other occurrences in this PR.

seedlit (commented):
looked over the surface.
Left a couple of comments.

up42/http/client.py
javidq: I'd like to address that in a follow up PR to include all benefits that 3.9 has over 3.8 in all source code

seedlit (approved):

PR #570, opened 2024-03-19
[AT-4915] Settings detection based on credentials

javidq:
Detection of credentials settings based on supplied credentials

Items:

For release:
* [x] Bumped version
* [x] Added changelog

### DIFF
--- a/up42/http/oauth.py
+++ b/up42/http/oauth.py
@@ -1,6 +1,7 @@
 import dataclasses as dc
 import datetime as dt
-from typing import Protocol
+import warnings
+from typing import Optional, Protocol
 
 import requests
 from requests import auth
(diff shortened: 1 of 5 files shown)


up42/http/oauth.py
```diff
@@ -93,3 +94,28 @@ def token(self) -> Token:
...
+        keys = credentials.keys()
+        if keys == {"project_id", "project_api_key"}:
+            warnings.warn(
+                "Project based authentication will be deprecated."
+                "Please follow authentication guidelines (/docs/authentication.md)."
+            )
+            return config.ProjectCredentialsSettings(**credentials)
+        if keys == {"username", "password"}:
+            return config.AccountCredentialsSettings(**credentials)
+        raise InvalidCredentials(f"Credentials {credentials} are not supported")
```
seedlit, line 110: just pointing out - this will expose the credentials. Even though the user themself would be aware of those credentials already, should we just expose the keys instead?

seedlit (commented):
left a couple of comments

up42/http/oauth.py
andher1802: In this case will expose the invalid credentials. But can be something to consider as the user can have a typo.

up42/http/oauth.py
```diff
@@ -93,3 +94,28 @@ def token(self) -> Token:
         if self._token.has_expired:
             self._token = self._fetch_token()
         return self._token
+
+
+def detect_settings(credentials: dict) -> Optional[config.CredentialsSettings]:
```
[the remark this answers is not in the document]
javidq: All those classes are used to hold and transfer state or to parameterize the behavior of a single function. In general this whole package is meant to be invoked using a single functional entry point.

seedlit (approved):

PR #572, opened 2024-03-20
[AT-4915] Switch to http client in auth.py

javidq:

Items:

For release:
* [x] Bumped version
* [x] Added changelog


up42/http/oauth.py
```diff
@@ -93,20 +93,21 @@ def token(self) -> Token:
...
-                "Project based authentication will be deprecated."
-                "Please follow authentication guidelines (/docs/authentication.md)."
-            )
-            return config.ProjectCredentialsSettings(**credentials)
-        if keys == {"username", "password"}:
-            return config.AccountCredentialsSettings(**credentials)
-        raise InvalidCredentials
-    elif any(credentials.values()):
-        raise IncompleteCredentials
+def detect_settings(credentials: Optional[dict]) -> Optional[config.CredentialsSettings]:
```
seedlit, line 96: are these changes coming from pre-commit hooks?
Is there a difference of version in your environment and poetry files?

seedlit (commented):
Nice!
left some comments.

up42/http/oauth.py
javidq: No, they are coming actual changes, config read from file can be None (I dropped or {} part since it also doesn't work with this code). Hence are the changes in the input signature and the actual code

seedlit (approved):

PR #575, opened 2024-03-27
[AT-4915] Dependency injection and test coverage for auth.py

javidq:
Injection of functional dependencies and test coverage for auth.py

Items:

For release:
* [x] Bumped version
* [x] Added changelog

### DIFF
--- a/up42/catalog.py
+++ b/up42/catalog.py
@@ -503,7 +503,7 @@ class Catalog(CatalogBase, viztools.VizTools):
                 with open(out_path, "wb") as dst:
                     for chunk in response:
                         dst.write(chunk)
-            except ValueError:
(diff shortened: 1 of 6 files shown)


up42/catalog.py
```diff
@@ -503,7 +503,7 @@ def download_quicklooks(
                 with open(out_path, "wb") as dst:
                     for chunk in response:
                         dst.write(chunk)
-            except ValueError:
+            except IOError:
```
javidq, line 506: It was getting before ValueError due to the bug fixed in auth.py in this PR - err.response.json() was failing due to invalid response content

seedlit (approved):
looked over the surface.
LGTM!

PR #577, opened 2024-04-02
[AT-4358] Use http session in auth.py

javidq:
- Added standard headers to `http/session.py`
- Added session provisioning to `auth.py` and `http/client.py`
- Dropped undocumented `authenticate` flag in `auth.py`
- Dropped undocumented kwargs in `auth.py` and `main.py`

Items:

For release:
* [x] Bumped version
* [x] Added changelog


up42/http/session.py
```diff
@@ -16,14 +19,19 @@ def request(self, *args, **kwargs) -> requests.Response:
...
     session = StatusValidatingSession()
     adapter = create_adapter()
     for schema in SCHEMAS:
         session.mount(schema + "://", adapter)
-    session.auth = create_auth(create_adapter=create_adapter)
+    session.auth = auth
+    session.headers = {
+        "Content-Type": "application/json",
+        "cache-control": "no-cache",
+        "User-Agent": f"up42-py/{version} ({REPOSITORY_URL})",
```
seedlit, line 35: do we need repo URL in the headers? I can't think of a use-case where we might be needing this.
If not, let's use this as an opportunity to drop it.

seedlit (commented):
left a comment.

up42/http/session.py
javidq: @seedlit I'm not sure whether this will affect our BI or not. Let's add this as a ticket for @PelJon to confirm first

seedlit (approved):

PR #581, opened 2024-04-04
[AT-4559] drop deprecated code - Visualization

seedlit:
NOTE: This is part 1 of multiple PRs I will be making (to make sure I don't end up creating too big of a PR).

Scope: Dropped Visualization functionality. Note that I have still kept the `Viztools` class as some other classes (for eg. Job) depends on it. I will drop it in subsequent PRs.

Following functions have been dropped:
folium_base_map(), plot_quicklooks(), plot_coverage(), draw_aoi(), _map_images() (internal function), map_quicklooks(), plot_coverage(), plot_results(), requires_viz() (internal function), map_results(), render() (internal function)

Items:

For release:
* [x] Bumped version
* [x] Added changelog


up42/__init__.py
```diff
@@ -56,7 +56,6 @@
 from up42.tasking import Tasking
 from up42.tools import get_example_aoi, read_vector_file
 from up42.utils import get_up42_py_version
-from up42.viztools import draw_aoi
```
javidq, line 59: We need to clean up mypy.ini and pylintrc as well

javidq (changes requested):

up42/viztools.py
```diff
@@ -2,617 +2,6 @@
 Visualization tools available in various objects
```
javidq, line 2: Why do we need to keep the module itself?

javidq (changes requested):

up42/viztools.py
seedlit: some other classes (for example [see this](https://github.com/up42/up42-py/blob/4c35515cab21e50ff8b8a233454e0f0036289508/up42/job.py#L18)), depend on class `VizTools`.
And I wanted to tackle it in multiple PRs.

up42/__init__.py
seedlit: good catch!

javidq (approved):

PR #587, opened 2024-04-15
[AT-4559] drop deprecated code part2 - Project; JobCollection

seedlit:
Jira ticket [here.](https://up42.atlassian.net/browse/AT-4559)

Scope:
- Dropped deprecated JobCollection functions - info, status, apply, download_results
- Dropped deprecated Project functions - info, get_workflows, get_project_settings, get_jobs
- Dropped Project's fixtures and tests
- Dropped JobCollection's fixtures and tests
- Dropped Workflow's get_jobs function


up42/project.py
```diff
@@ -1,155 +0,0 @@
-import logging
-from typing import Dict, List, Union
-
-from up42.auth import Auth
-from up42.host import endpoint
-from up42.job import Job
-from up42.jobcollection import JobCollection
-from up42.utils import filter_jobs_on_mode, get_logger
```
javidq, line 8: filter_jobs_on_mode is not used anymore

javidq (changes requested):

up42/project.py
seedlit: good catch!

javidq (approved):

PR #589, opened 2024-04-15
[AT-4559] drop deprecated code part4 - JobTask

seedlit:
Jira ticket [here.](https://up42.atlassian.net/browse/AT-4559)

Scope:
- Dropped deprecated Jobtask functions - info, get_results_json, download_results, download_quicklooks
- Dropped Jobtask tests and fixtures


tests/fixtures/fixtures_globals.py
```diff
@@ -46,8 +46,6 @@
 JOB_ID = "13ba070a-55b7-4b3f-95f1-7b11ac8f1175"
```
javidq, line 46: These two and the related code can be dropped as well

javidq (approved):

PR #607, opened 2024-05-15
PT-5266 Run sdk-tests in parallel

janchrizz:
Updating the run command for `sdk-tests`, so tests run with 10 workers in parallel, only merge after this https://github.com/up42/sdk-tests/pull/51 is merged!


andher1802 (approved):
LGTM

PR #615, opened 2024-05-27
PT-5054 Adding session and workspace descriptors

andher1802:
This PR adds `Session` and `WorkspaceId` descriptors to main.py and adds the correspondent tests. 

Items:

For release:
* [x] Bumped version
* [x] Added changelog


tests/test_main.py
```diff
@@ -32,6 +32,7 @@ def workspace(self, auth_mock):
             workspace_mock.auth = auth_mock
             workspace_mock.id = constants.WORKSPACE_ID
             yield
+        main.workspace = main._Workspace()  # pylint: disable=protected-access
```
andher1802, line 35: The fixture `workspace` causes an interaction with other tests outside the TestClass. This means that if I run tests individually, they are fine. However, if I run tests altogether, there is an interaction due to the workspace singleton, and causes the `test_should_fail_if_not_authenticated` session tests to fail when run together with other tests.

For this reason, I needed to reset the fixture to the initial condition to make sure it did not affect other tests. This is not super elegant, but it solved the problem. Another solution I tested was to set the scope of the workspace fixture to "function" but I think the state of the singleton is not reset.

tests/test_main.py
```diff
@@ -64,3 +65,55 @@ def test_get_credits_balance(self, requests_mock):
...
+            workspace_id = main.WorkspaceId()
+
+            def __init__(self, workspace_id=None):
+                if workspace_id:
+                    self.__dict__["workspace_id"] = workspace_id
+
+        instance_with_id = DummyInstance(workspace_id="custom_id")
+        assert instance_with_id.workspace_id == "custom_id"
+
+    def test_should_provide_workspace_id_if_not_provided(self, dummy_instance_without_local_id):
```
andher1802, line 118: Here, I think there is still one branch not tested.

CHANGELOG.md
```diff
@@ -29,6 +29,14 @@ You can check your current version with the following command:
     ```
 
+
+## 1.0.4a3
+
+**May 27, 2024**
+
+- Added `Session` and `WorkspaceID` descriptors to provide access to session and workspace_id from other Classes.
```
javidq, line 37: ID -> Id, Classes -> classes

up42/main.py
```diff
@@ -3,6 +3,8 @@
 import warnings
```
javidq, line 3: this file to be renamed to base

tests/test_main.py
javidq: It's unrelated, the reason is because main is authenticated at this point, therefore you need to patch it workspace down below in the class that fails similar how it is done here.

tests/test_main.py
```diff
@@ -64,3 +65,64 @@ def test_get_credits_balance(self, requests_mock):
             json={"data": balance},
         )
         assert main.get_credits_balance() == balance
+
+
+class TestSession:
+    @pytest.fixture(scope="function")
```
javidq, line 71: scope is not needed

tests/test_main.py
```diff
@@ -64,3 +65,64 @@ def test_get_credits_balance(self, requests_mock):
             json={"data": balance},
         )
         assert main.get_credits_balance() == balance
+
+
+class TestSession:
```
javidq, line 70: workspace mock fixture similar to the one above is needed here

tests/test_main.py
```diff
@@ -64,3 +65,64 @@ def test_get_credits_balance(self, requests_mock):
...
+            session = main.Session()
+
+        return DummyInstance()
+
+    def test_should_fail_if_not_authenticated(self, dummy_instance):
+        with pytest.raises(main.UserNotAuthenticated):
+            _ = dummy_instance.session
+
+    def test_should_provide_session_if_authenticated(self, requests_mock, dummy_instance):
+        requests_mock.post("https://api.up42.com/oauth/token", json={"access_token": constants.TOKEN})
```
javidq, line 83: Must rely on workspace_mock

tests/test_main.py
```diff
@@ -64,3 +65,64 @@ def test_get_credits_balance(self, requests_mock):
             json={"data": balance},
         )
         assert main.get_credits_balance() == balance
+
+
+class TestSession:
+    @pytest.fixture(scope="function")
+    def dummy_instance(self):
+        class DummyInstance:
```
javidq, line 73: class ActiveRecord and therefore record as instance name

tests/test_main.py
```diff
@@ -64,3 +65,64 @@ def test_get_credits_balance(self, requests_mock):
...
+        yield DummyInstance()
+        main.workspace = main._Workspace()  # pylint: disable=protected-access
+
+    def test_should_provided_local_if_set_workspace(self):
+        class DummyInstance:
+            workspace_id = main.WorkspaceId()
+
+            def __init__(self, workspace_id=None):
+                if workspace_id:
+                    self.__dict__["workspace_id"] = workspace_id
```
javidq, line 114: self.workspace_id = workspace_id

javidq: And the class can be used in all cases

tests/test_main.py
javidq: And the same below

tests/test_main.py
javidq: And I'd recommend to just use a data class with eq=True and check whether it affects equality check

javidq (changes requested):

PR #618, opened 2024-05-29
[PT-5013] Webhook active record

javidq:
First part of active record transformation of webhooks

Items:

For release:
* [x] Bumped version
* [x] Added changelog

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -38,8 +38,8 @@ class TestWorkspace:
 @dataclasses.dataclass(eq=True)
 class ActiveRecord:
     session = base.Session()
-    class_workspace_id = base.ClassWorkspaceId()
-    workspace_id: Union[str, base.InstanceWorkspaceId] = dataclasses.field(default=base.InstanceWorkspaceId())
+    class_workspace_id = base.WorkspaceId()
+    workspace_id: Union[str, base.WorkspaceId] = dataclasses.field(default=base.WorkspaceId())
 
 
 class TestDescriptors:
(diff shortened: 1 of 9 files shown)


tests/test_base.py
```diff
@@ -38,6 +38,7 @@ def test_should_get_credits_balance(self, requests_mock):
 @dataclasses.dataclass(eq=True)
 class ActiveRecord:
     session = base.Session()
+    class_workspace_id = base.WorkspaceId()
```
andher1802, line 41: I get it. We should change the workspace_id of the class only, while the global workspace_id shock remain the account workspace_id

andher1802 (approved):
Approved, nice job

tests/test_base.py
javidq: It's rather two scenarios merged into one.
1. In most of the cases (e.g. asset) a user deals with objects using different workspace ids
2. webhook is a special case, since a user always uses the same workspace id for all operations.

javidq: At least this is how it was for the current approach in SDK

up42/base.py
```diff
@@ -84,7 +84,9 @@ def __get__(self, obj, obj_type=None) -> requests.Session:
 
 class WorkspaceId:
     def __get__(self, obj, obj_type=None) -> str:
-        return obj.__dict__.get("workspace_id", workspace.id)
```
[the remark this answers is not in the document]
javidq: See the comment below

PR #630, opened 2024-06-11
[PT-4557] Query jobs

javidq:
Query jobs via iterator

Items:

For release:
* [x] Bumped version
* [x] Added changelog

### DIFF
--- a/up42/processing.py
+++ b/up42/processing.py
@@ -49,6 +49,10 @@ class JobSorting:
     credits = utils.SortingField("creditConsumption.credits")
 
 
+def _to_datetime(value: Optional[str]):
+    return value and datetime.datetime.fromisoformat(value.rstrip("Z"))
+
+
 @dataclasses.dataclass
 class Job:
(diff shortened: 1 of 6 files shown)


up42/processing.py
```diff
@@ -42,6 +42,17 @@ class JobMetadata(TypedDict):
...
 
 
+class JobSorting:
+    process_id = utils.SortingField("processID")
+    status = utils.SortingField("status")
+    created = utils.SortingField("created")
+    credits = utils.SortingField("creditConsumption.credits")
+
+
+def _to_datetime(value: Optional[str]):
```
andher1802, line 52: As we moved out this method from the `Job` class, and there is a method in `utils` named `format_time`. Should we move this to utils and group with the format_time method?

andher1802 (approved):
LGTM

up42/processing.py
javidq: Only if we see reuse

PR #634, opened 2024-06-13
[PT-5365] Errors and credits in processing jobs

javidq:
- Errors and credits retrieval in processing jobs
- Dropped documentation update since all documentation has moved to sdk.up42.com

Items:

For release:
* [x] Bumped version
* [x] Added changelog


up42/processing.py
```diff
@@ -83,12 +88,18 @@ def collection(self) -> Optional[pystac.Collection]:
 
     @staticmethod
     def from_metadata(metadata: JobMetadata) -> "Job":
+        results: JobResults = metadata.get("results") or {}
+        errors = results.get("errors") or []
+        validation_errors = [ValidationError(**error) for error in errors]
+        consumption = metadata.get("creditConsumption") or {}
```
andher1802, line 94: should this be _consumption if not included in tests

up42/processing.py
```diff
@@ -71,6 +74,8 @@ class Job:
     created: datetime.datetime
     updated: datetime.datetime
     collection_url: Optional[str] = None
+    errors: Optional[List[ValidationError]] = None
+    credits: Optional[int] = None
```
andher1802, line 78: I see in the job template we create a class for comparing JobTemplate costs. Why did we choose to just use credits in this case?. `hold_id` and `consumption_id` are not explicit properties with this approach. If that is intended then OK with me.

tests/test_processing.py
```diff
@@ -49,13 +49,18 @@
...
     "jobID": JOB_ID,
     "accountID": ACCOUNT_ID,
     "workspaceID": constants.WORKSPACE_ID,
     "definition": DEFINITION,
-    "results": {"collection": f"{COLLECTION_URL}"},
+    "results": {
+        "collection": f"{COLLECTION_URL}",
+        "errors": [dataclasses.asdict(INVALID_TITLE_ERROR)],
+    },
+    "creditConsumption": {"credits": CREDITS},
```
andher1802, line 63: `hold_id` and `consumption_id` are not included. It is OK, but just a remark considering that in the JobTemplate we created a class for JobCost. I was wondering why we chose a different approach here.

andher1802 (approved):
LGTM, just a one question regarding the generic dict approach for credit_consumption.

PR #646, opened 2024-06-26
[PT-4506] Improve OAuth flow

javidq:
See changelog

Checklist:
* [x] Test coverage
* [x] Bumped version
* [x] Added changelog

### DIFF
--- a/tests/test_host.py
+++ b/tests/test_host.py
@@ -7,3 +7,9 @@ from up42 import host
 def test_should_use_domain_in_endpoint_url(domain):
     host.DOMAIN = domain
     assert host.endpoint("/path") == f"https://api.up42.{domain}/path"
+
+
+@pytest.mark.parametrize("domain", ["xyz", "abc"])
+def test_should_use_domain_in_token_endpoint_url(domain):
+    host.DOMAIN = domain
+    assert domain in host.token_endpoint()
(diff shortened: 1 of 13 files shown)


up42/host.py
```diff
@@ -4,3 +4,7 @@
 def endpoint(path: str):
     """Gets endpoint url based on its path."""
     return f"https://api.up42.{DOMAIN}{path}"
+
+
+def token_endpoint():
+    return f"https://auth.up42.{DOMAIN}/realms/public/protocol/openid-connect/token"
```
JeremiahDo, line 10:
```suggestion
    return f"https://auth.up42.com/realms/public/protocol/openid-connect/token"
```
NIT Why do we need `DOMAIN` at all? I see it always being `com`. If we want to keep a constant we could also make it `https://auth.up42.com`.

tests/test_host.py
```diff
@@ -7,3 +7,9 @@
 def test_should_use_domain_in_endpoint_url(domain):
     host.DOMAIN = domain
     assert host.endpoint("/path") == f"https://api.up42.{domain}/path"
+
+
+@pytest.mark.parametrize("domain", ["xyz", "abc"])
+def test_should_use_domain_in_token_endpoint_url(domain):
```
JeremiahDo, line 13: Why would we test the domain only? If anything it would make more sense to me to test that the entire URL. And in the end we're testing it's right by using the fixture `fixture_globals.TOKEN_ENDPOINT`.

JeremiahDo (changes requested):

JeremiahDo (approved):

jmigueldelgado (approved):

PR #648, opened 2024-07-09
Clean up product glossary usage

javidq:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,16 @@ You can check your current version with the following command:
     ```
 
 For more information, see [UP42 Python package description](https://pypi.org/project/up42-py/).
+## 1.1.1a4
+
+**Jul 15, 2024**
+- Extract `ProductGlossary` in `catalog.py`.
+- Provide type hints for `ProductGlossary` methods.
+- Simplify the usage of `get_data_products` in `CatalogBase`.
+- Fix type hint for `get_webhook_events`.
+- Reduce the usage `auth::request` in `base.py`.
+
+
 ## 1.1.1a3
 
 **Jul 8, 2024**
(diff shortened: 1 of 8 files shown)


up42/catalog.py
```diff
@@ -4,84 +4,179 @@
...
+    description: str
+    createdAt: Optional[str]
+    updatedAt: Optional[str]
+    isIntegrated: bool
+
+
+CollectionType: TypeAlias = Literal["ARCHIVE", "TASKING"]
+
+
+class Collection(TypedDict):
```
andher1802, line 50: I know this is something that we cant change because it comes from the endpoint. But we are adding an extra layer of confussing our users by naming collections in the catalog. We need to explain not only that STAC collection are assets but now that we have catalog collections that correspond to the providers datasets. In the ideal STAC world STAC Collection should correspond to Provider Collection.

andher1802: Perhaps some naming or documentation at the SDK level that we can help better the users to understand this.

up42/catalog.py
```diff
@@ -4,84 +4,179 @@
...
+        self.auth = auth
+        self.workspace_id = workspace_id
+        # FIXME: cannot be optional
+        self.type: Optional[CollectionType] = None
+
+    def get_collections(self) -> list[Collection]:
+        return ProductGlossary.get_collections(self.type)
 
-            return collection_overview
+    def get_data_products(self, basic: bool = True) -> Union[List[DataProduct], dict[str, CollectionOverview]]:
```
andher1802, line 178: Should we add here the deprecation message for V2 in the near future?

CHANGELOG.md
```diff
@@ -29,6 +29,16 @@ You can check your current version with the following command:
     ```
 
+## 1.1.1a4
+
+**Jul 15, 2024**
```
andher1802, line 34: 15? are we aiming for Monday

andher1802 (approved):

PR #651, opened 2024-07-15
[PT-5447] Improve coverage of catalog quicklooks download

javidq:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/up42/auth.py
+++ b/up42/auth.py
@@ -93,7 +93,6 @@ class Auth:
         request_type: str,
         url: str,
         data: dict = {},
-        return_text: bool = True,
     ):
         """
         Makes a request to the API, handles authentication and SDK headers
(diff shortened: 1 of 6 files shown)


up42/auth.py
```diff
@@ -109,22 +108,15 @@ def request(
...
-                        and response_json.get("data") is None
-                    ):
-                        raise ValueError(response_json["error"])
-                    # Catalog search, JobTask logs etc. does not have the usual {"data": {}, "error": {}} format.
-                    return response_json
-                except json.JSONDecodeError:  # e.g. JobTask logs are str format.
-                    return response.text
-            else:  # E.g. for DELETE
-                return response
+            try:
```
andher1802, line 111: Just a question: 
Why do we need two nested try statements? 
Cannot we have one and multiple `except` depending on the error?

andher1802 (approved):

up42/auth.py
javidq: The variable used in one of the except blocks is not available at all in the other one. Python does allow that, but it hurts ease of understanding. These methods will be gone soon anyway

PR #655, opened 2024-07-29
[PT-5519] EULA acceptance check for processing templates

javidq:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/up42/processing.py
+++ b/up42/processing.py
@@ -21,6 +21,8 @@ class ValidationError:
 
 class JobStatus(enum.Enum):
     CREATED = "created"
+    LICENSED = "licensed"
+    UNLICENSED = "unlicensed"
     VALID = "valid"
     INVALID = "invalid"
     ACCEPTED = "accepted"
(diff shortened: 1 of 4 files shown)


up42/processing.py
```diff
@@ -131,7 +131,7 @@ def update():
             self.collection_url = job.collection_url
             self.errors = job.errors
             self.credits = job.credits
-            if self.status not in [JobStatus.CAPTURED, JobStatus.RELEASED]:
+            if self.status not in [JobStatus.CAPTURED, JobStatus.RELEASED, JobStatus.INVALID, JobStatus.REJECTED]:
```
matheuspinheirosantos, line 134: Should we also check for the `unlicensed` state? Even though it should not be possible to have it when starting a job from the SDK?

javidq: Added

andher1802 (approved):
LGTM

PR #667, opened 2024-08-23
PT-5591 Order class to descriptor

andher1802:
See changelog.

Checklist:
* [ ] Test coverage
* [ ] Version bump
* [ ] Changelog update

### DIFF
--- a/tests/test_order.py
+++ b/tests/test_order.py
@@ -1,9 +1,20 @@
+from unittest import mock
+
 import pytest
+import requests
 
 from up42 import asset, order
 
 from .fixtures import fixtures_globals as constants
 from .fixtures import fixtures_order
+
+
+@pytest.fixture(autouse=True)
+def workspace():
+    with mock.patch("up42.base.workspace") as workspace_mock:
+        workspace_mock.auth.session = requests.session()
+        workspace_mock.id = constants.WORKSPACE_ID
+        yield
 
 
 def test_init(order_mock):
@@ -16,14 +27,14 @@ def test_order_info(order_mock):
     assert order_mock.info["id"] == constants.ORDER_ID
 
 
-def test_repr(auth_mock):
+def test_repr():
     order_info = {
         "id": "your_order_id",
         "status": "PLACED",
         "createdAt": "2023-01-01T12:00:00Z",
         "updatedAt": "2023-01-01T12:30:00Z",
     }
-    order_placed = order.Order(auth_mock, order_id="your_order_id", order_info=order_info)
+    order_placed = order.Order(order_id="your_order_id", order_info=order_info)
 
     expected_repr = (
         "Order(order_id: your_order_id, status: PLACED,"
@@ -67,15 +78,22 @@ def test_order_parameters(order_mock):
     assert not order_mock.order_parameters
 
 
-def test_get_assets_should_search_assets_by_order_id(auth_mock, requests_mock):
-    order_response = {"id": constants.ORDER_ID, "status": "FULFILLED"}
+@pytest.mark.parametrize(
+    "status",
+    [
+        "FULFILLED",
+        "BEING_FULFILLED",
+    ],
+)
+def test_get_assets_should_search_assets_by_order_id(requests_mock, status):
+    order_response = {"id": constants.ORDER_ID, "status": status}
 
     url_order_info = f"{constants.API_HOST}/v2/orders/{constants.ORDER_ID}"
     requests_mock.get(url=url_order_info, json=order_response)
-    url_asset_info = f"{constants.API_HOST}/v2/assets?search={constants.ORDER_ID}&size=50"
+    url_asset_info = f"{constants.API_HOST}/v2/assets?search={constants.ORDER_ID}&page=0"
     requests_mock.get(url=url_asset_info, json=fixtures_order.JSON_GET_ASSETS_RESPONSE)
-    order_placed = order.Order(auth=auth_mock, order_id=constants.ORDER_ID)
-    (asset_returned,) = order_placed.get_assets()
+    order_placed = order.Order(order_id=constants.ORDER_ID)
+    (asset_returned,) = list(order_placed.get_assets())
     assert isinstance(asset_returned, asset.Asset)
     assert asset_returned.asset_id == constants.ASSET_ORDER_ID
 
@@ -88,19 +106,18 @@ def test_get_assets_should_search_assets_by_order_id(auth_mock, requests_mock):
         "PLACED",
         "PLACEMENT_FAILED",
         "DELIVERY_INITIALIZATION_FAILED",
-        "BEING_FULFILLED",
         "DOWNLOAD_FAILED",
         "DOWNLOADED",
         "FAILED_PERMANENTLY",
     ],
 )
-def test_should_fail_to_get_assets_for_unfulfilled_order(auth_mock, requests_mock, status):
+def test_should_fail_to_get_assets_for_unfulfilled_order(requests_mock, status):
     order_response = {"id": constants.ORDER_ID, "status": status}
     url_order_info = f"{constants.API_HOST}/v2/orders/{constants.ORDER_ID}"
     requests_mock.get(url=url_order_info, json=order_response)
-    order_placed = order.Order(auth=auth_mock, order_id=constants.ORDER_ID)
+    order_placed = order.Order(order_id=constants.ORDER_ID)
     with pytest.raises(ValueError):
-        order_placed.get_assets()
+        list(order_placed.get_assets())
 
 
 def test_place_order(catalog_order_parameters, auth_mock, order_mock, requests_mock):
@@ -111,7 +128,7 @@ def test_place_order(catalog_order_parameters, auth_mock, order_mock, requests_m
             "errors": [],
         },
     )
-    order_placed = order.Order.place(auth_mock, catalog_order_parameters, constants.WORKSPACE_ID)
+    order_placed = order.place(auth_mock, catalog_order_parameters, constants.WORKSPACE_ID)
     assert order_placed == order_mock
     assert order_placed.order_id == constants.ORDER_ID
     assert order_placed.order_parameters == catalog_order_parameters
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,15 @@ You can check your current version with the following command:
...
 For more information, see [UP42 Python package description](https://pypi.org/project/up42-py/).
 
+## 2.1.0a1
+
+**TBD, 2024**
+
+- Switching `Order` class to descriptor
+- `Order.place` and `Order.estimate` moved to `place_order` and `estimate_order` module level functions.
+- `Order.get_assets` now allows to get assets from orders in `BEING_FULFILLED` state.
+- `Order.get_assets` now returning an iterator instear of list.
```
andher1802, line 40: Potential breaking change, reconsider version to major version update.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "up42-py"
-version = "2.0.1"
+version = "2.1.0a1"
```
andher1802, line 3: see my changelog comment

up42/order.py
```diff
@@ -118,68 +196,16 @@ def is_fulfilled(self) -> bool:
...
-            estimated_credits,
-            order_parameters,
-        )
-        return estimated_credits
+        if not (self.is_fulfilled) and self.status != "BEING_FULFILLED":
+            raise ValueError(f"Order {self.order_id} is not valid. Current status is {self.status}")
+        params = {"search": self.order_id, "page": 0}
+        for page in self._get_pages(self.session, "/v2/assets", params):
+            for asset_info in page:
+                yield asset.Asset(base.workspace.auth, asset_info=asset_info)
```
andher1802, line 208: we can't completely get rid of auth element until we migrate asset class to descriptor.

CHANGELOG.md
```diff
@@ -30,6 +30,12 @@ You can check your current version with the following command:
 
 For more information, see [UP42 Python package description](https://pypi.org/project/up42-py/).
 
+## 2.1.0a1
+
+**Aug 26, 2024**
+- `Order.get_assets` now allows to get assets from orders in `BEING_FULFILLED` state.
+- `Order.get_assets` now returning an iterator instear of list.
```
andher1802, line 37: Can be this considered a breaking change, I think so. Should we bump to major version bump or force the output to list instead of Iterator.

tests/test_order.py
```diff
@@ -127,13 +144,11 @@ def test_place_order_fails_if_response_contains_error(catalog_order_parameters,
         },
     )
     with pytest.raises(ValueError) as err:
-        order.Order.place(auth_mock, catalog_order_parameters, constants.WORKSPACE_ID)
+        order.Order.place(catalog_order_parameters, constants.WORKSPACE_ID)
     assert error_content in str(err.value)
 
 
 def test_track_status_running(order_mock, requests_mock):
-    del order_mock._info
```
andher1802, line 135: This was never needed IMO.

tests/test_order.py
```diff
@@ -160,14 +180,13 @@ def test_track_status_running(order_mock, requests_mock):
         },
     ]
     requests_mock.get(url_job_info, status_responses)
-    order_status = order_mock.track_status(report_time=0.1)
+    order_test = order.Order(constants.ORDER_ID)
```
andher1802, line 183: This was to fix an error in tests. Previously the test called the endpoint by the order_mock and the last of the status_responses, now the test is having the expected behavior.

CHANGELOG.md
javidq: We should not make breaking change at this point, let's keep the signature

tests/test_order.py
```diff
@@ -1,11 +1,24 @@
...
+from typing import Any, Dict
+from unittest import mock
+
 import pytest
+import requests
 
 from up42 import asset, order
 
 from .fixtures import fixtures_globals as constants
 from .fixtures import fixtures_order
```
javidq, line 12: We aligned to fix the coverage. It means we need to revisit all tests in this module, not only the ones affected

javidq (changes requested):

up42/catalog.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, auth: up42_auth.Auth, workspace_id: str):
         super().__init__(auth, workspace_id)
         self.type: glossary.CollectionType = glossary.CollectionType.ARCHIVE
 
-    def estimate_order(self, order_parameters: Optional[Dict]) -> int:
+    def estimate_order(self, order_parameters: order.OrderParams) -> int:
```
andher1802, line 98: same here, there is not optional value for `order_params` in `estimate_order`

tests/test_order.py
```diff
@@ -95,12 +107,11 @@ def test_should_get_assets_if_fulfilled(self, auth_mock, requests_mock: req_mock
...
             ],
-            "totalPages": 1,
-            "totalElements": 1,
+            "totalPages": 2,
         }
         requests_mock.get(url=url_asset_info, json=asset_info)
-        order_obj = order.Order(auth=auth_mock, order_id=constants.ORDER_ID)
-        (asset_obj,) = order_obj.get_assets()
+        order_obj = order.Order(order_id=constants.ORDER_ID)
+        asset_obj = order_obj.get_assets()[0]
```
javidq, line 114: Let's keep the previous approach here, it verifies also that there's only one item there

up42/catalog.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, auth: up42_auth.Auth, workspace_id: str):
...
         """
         Estimate the cost of an order.
```
javidq, line 100: We don't need type:ignore below

up42/order.py
```diff
@@ -28,10 +59,31 @@ def _translate_construct_parameters(order_parameters):
...
+    "BEING_FULFILLED",
+    "DELIVERY_INITIALIZATION_FAILED",
+    "DOWNLOAD_FAILED",
+    "DOWNLOADED",
+    "FULFILLED",
+    "FAILED",
+    "FAILED_PERMANENTLY",
+]
+
+OrderSubtatus = Literal[
```
javidq, line 78: Not needed

up42/order.py
```diff
@@ -1,25 +1,56 @@
...
+    """
+
+    dataProduct: str  # pylint: disable=invalid-name
+    params: Dict[str, Any]
+    tags: List[str]
+
+
+class OrderParamsV2(TypedDict, total=False):
+    """
+    Represents the stucture data format for the order parameters for the V2 endpoint.
```
javidq, line 31: the stucture data format -> schema

up42/order.py
```diff
@@ -180,7 +230,7 @@ def estimate(auth: up42_auth.Auth, order_parameters: OrderParams) -> int:
         response_json = auth.request(
```
javidq, line 230: We don't need auth here as well

javidq: Let's convert this to class method to use cls.session

javidq: And no casting is needed

up42/order.py
```diff
@@ -131,14 +168,31 @@ def get_assets(self) -> List[asset.Asset]:
...
         Gets the Order assets or results.
         """
-        if self.is_fulfilled:
-            params: asset_searcher.AssetSearchParams = {"search": self.order_id}
-            assets_response = asset_searcher.search_assets(self.auth, params=params)
-            return [asset.Asset(asset_info=asset_info) for asset_info in assets_response]
-        raise UnfulfilledOrder(f"Order {self.order_id} is not FULFILLED! Current status is {self.status}")
+        current_info = copy.deepcopy(self._info)
+
+        def get_pages(endpoint: str, params: dict[str, Any]):
```
javidq, line 173: Let's simplify this to
```python
        def get_pages(endpoint: str):
            while True:
                response = self.session.get(
                    host.endpoint(endpoint), params=params
                ).json()
                yield response["content"]
                params["page"] += 1
                if params["page"] == response["totalPages"]:
                    break
```
and define after params

up42/order.py
```diff
@@ -131,14 +168,31 @@ def get_assets(self) -> List[asset.Asset]:
...
+            response = self.session.get(host.endpoint(endpoint), params=params).json()
+            total_pages = response["totalPages"]
+            while True:
+                yield response["content"]
+                params["page"] += 1
+                if params["page"] == total_pages:
+                    break
+                response = self.session.get(host.endpoint(endpoint), params=params).json()
+
+        if current_info["status"] != "FULFILLED" and current_info["status"] != "BEING_FULFILLED":
```
javidq, line 183: Let's simplify this to 
```python
        if (status := current_info["status"]) not in ["FULFILLED", "BEING_FULFILLED"]:
            raise UnfulfilledOrder(
                f"""Order {self.order_id} is not valid. Current status is {status}"""
            )
```

tests/test_order.py
```diff
@@ -130,22 +140,22 @@ def test_fails_to_get_assets_if_not_fulfilled(self, auth_mock, requests_mock: re
         [{"type": "ARCHIVE"}, {"type": "TASKING", "orderDetails": {"subStatus": "substatus"}}],
         ids=["ARCHIVE", "TASKING"],
     )
-    def test_should_track_order_status_until_fulfilled(self, auth_mock, requests_mock: req_mock.Mocker, info: dict):
+    def test_should_track_order_status_until_fulfilled(self, requests_mock: req_mock.Mocker, info: dict):
         responses = []
```
javidq, line 144: Not needed

CHANGELOG.md
```diff
@@ -30,6 +30,15 @@ You can check your current version with the following command:
 
 For more information, see [UP42 Python package description](https://pypi.org/project/up42-py/).
 
+## 2.1.0a4
+
+**Sep 11, 2024**
+- `Order.get_assets` now allows to get assets from orders in `BEING_FULFILLED` state.
+- Switch `Orders` class to use `session` descriptor.
```
javidq, line 37: Order

CHANGELOG.md
```diff
@@ -30,6 +30,15 @@ You can check your current version with the following command:
...
+- Set `OrderParams` as input type for `CatalogBase::place` and `Catalog::estimate` from `Optional[dict]`.
+- Add `OrderParamsV2` as output type for `_translate_construct_parameters` in order module.
```
javidq, line 39: Added

CHANGELOG.md
```diff
@@ -30,6 +30,15 @@ You can check your current version with the following command:
...
+- Switch `Order::status` from `str` to `Literal`.
```
javidq, line 40: Changed `Order::status` type

javidq (changes requested):

up42/order.py
javidq: Unaddressed

tests/test_order.py
javidq: That's exactly what we're trying to test. If for some reason it would return more than expected this test must fail

javidq (approved):

PR #669, opened 2024-08-27
Improve order test coverage

javidq:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/tests/test_order.py
+++ b/tests/test_order.py
@@ -114,7 +114,7 @@ class TestOrder:
         info = {"status": status}
         requests_mock.get(url=ORDER_URL, json=info)
         order_obj = order.Order(auth=auth_mock, order_id=constants.ORDER_ID, order_info=info)
-        with pytest.raises(order.OrderUnfulfilledError) as err:
+        with pytest.raises(order.UnfulfilledOrder) as err:
             order_obj.get_assets()
         assert re.compile(f".*{constants.ORDER_ID}.*{status}").match(str(err.value)) is not None
 
--- a/tests/test_order.py
+++ b/tests/test_order.py
@@ -1,35 +1,69 @@
+from unittest import mock
+
 import pytest
+import requests_mock as req_mock
 
 from up42 import asset, order
-
 from .fixtures import fixtures_globals as constants
 from .fixtures import fixtures_order
 
 
-def test_init(order_mock):
-    assert isinstance(order_mock, order.Order)
-    assert order_mock.order_id == constants.ORDER_ID
+class TestOrder:
+    def test_should_initialize_with_info_provided(self):
+        auth = mock.MagicMock()
+        info = {"some": "data"}
+        order_obj = order.Order(auth, constants.ORDER_ID, info)
+        assert order_obj.auth == auth
+        assert order_obj.order_id == constants.ORDER_ID
+        assert order_obj._info == info
 
+    def test_should_initialize(self, auth_mock, requests_mock: req_mock.Mocker):
+        url = f"{constants.API_HOST}/v2/orders/{constants.ORDER_ID}"
+        info = {"some": "data"}
+        requests_mock.get(url=url, json=info)
+        order_obj = order.Order(auth_mock, constants.ORDER_ID)
+        assert order_obj.auth == auth_mock
+        assert order_obj.order_id == constants.ORDER_ID
+        assert order_obj.info == info
 
-def test_order_info(order_mock):
-    assert order_mock.info
-    assert order_mock.info["id"] == constants.ORDER_ID
+    def test_should_provide_representation(self, auth_mock):
+        info = {
+            "id": constants.ORDER_ID,
+            "status": "PLACED",
+            "createdAt": "2023-01-01T12:00:00Z",
+            "updatedAt": "2023-01-01T12:30:00Z",
+        }
+        order_placed = order.Order(
+            auth_mock, order_id=constants.ORDER_ID, order_info=info
+        )
 
+        expected_repr = (
+            f"Order(order_id: {constants.ORDER_ID}, status: PLACED,"
+            "createdAt: 2023-01-01T12:00:00Z, updatedAt: 2023-01-01T12:30:00Z)"
+        )
+        assert repr(order_placed) == expected_repr
 
-def test_repr(auth_mock):
-    order_info = {
-        "id": "your_order_id",
-        "status": "PLACED",
-        "createdAt": "2023-01-01T12:00:00Z",
-        "updatedAt": "2023-01-01T12:30:00Z",
-    }
-    order_placed = order.Order(auth_mock, order_id="your_order_id", order_info=order_info)
+    def test_should_provide_status(self):
+        info = {"status": "random"}
+        pass
 
-    expected_repr = (
-        "Order(order_id: your_order_id, status: PLACED,"
-        "createdAt: 2023-01-01T12:00:00Z, updatedAt: 2023-01-01T12:30:00Z)"
+    @pytest.mark.parametrize(
+        "info, expected",
+        [
+            ({"type": "ARCHIVE"}, {}),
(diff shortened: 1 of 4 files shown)


tests/test_order.py
```diff
@@ -1,106 +1,112 @@
...
-    with pytest.raises(ValueError):
-        order_placed.get_assets()
+    def test_should_compute_is_fulfilled(self, auth_mock, requests_mock: req_mock.Mocker, info: dict, expected):
+        requests_mock.get(url=ORDER_URL, json=info)
+        order_obj = order.Order(auth_mock, order_id=constants.ORDER_ID, order_info=info)
+        assert order_obj.is_fulfilled == expected
+
+    def test_get_assets_should_search_assets_by_order_id(self, auth_mock, requests_mock):
+        order_response = {"id": constants.ORDER_ID, "status": "FULFILLED"}
+        requests_mock.get(url=ORDER_URL, json=order_response)
```
javidq, line 82: We don't need to mock this if we're passing it in the constructor

tests/test_order.py
```diff
@@ -1,106 +1,112 @@
...
-    order_placed = order.Order(auth=auth_mock, order_id=constants.ORDER_ID)
-    with pytest.raises(ValueError):
-        order_placed.get_assets()
+    def test_should_compute_is_fulfilled(self, auth_mock, requests_mock: req_mock.Mocker, info: dict, expected):
+        requests_mock.get(url=ORDER_URL, json=info)
+        order_obj = order.Order(auth_mock, order_id=constants.ORDER_ID, order_info=info)
+        assert order_obj.is_fulfilled == expected
+
+    def test_get_assets_should_search_assets_by_order_id(self, auth_mock, requests_mock):
+        order_response = {"id": constants.ORDER_ID, "status": "FULFILLED"}
```
javidq, line 81: info

tests/test_order.py
```diff
@@ -1,106 +1,112 @@
...
+        requests_mock.get(url=ORDER_URL, json=order_response)
+        url_asset_info = f"{constants.API_HOST}/v2/assets?search={constants.ORDER_ID}"
+        requests_mock.get(url=url_asset_info, json=fixtures_order.JSON_GET_ASSETS_RESPONSE)
```
javidq, line 84: Let's not use legacy responses here, we need a minimal single asset payload

tests/test_order.py
```diff
@@ -1,106 +1,112 @@
...
+        order_obj = order.Order(auth=auth_mock, order_id=constants.ORDER_ID, order_info=order_response)
+        asset_returned = order_obj.get_assets()
```
javidq, line 86: assets

tests/test_order.py
```diff
@@ -1,106 +1,112 @@
...
-    requests_mock.get(url=url_order_info, json=order_response)
-    order_placed = order.Order(auth=auth_mock, order_id=constants.ORDER_ID)
-    with pytest.raises(ValueError):
-        order_placed.get_assets()
+    def test_should_compute_is_fulfilled(self, auth_mock, requests_mock: req_mock.Mocker, info: dict, expected):
+        requests_mock.get(url=ORDER_URL, json=info)
+        order_obj = order.Order(auth_mock, order_id=constants.ORDER_ID, order_info=info)
+        assert order_obj.is_fulfilled == expected
+
+    def test_get_assets_should_search_assets_by_order_id(self, auth_mock, requests_mock):
```
javidq, line 80: test_should_get_assets_if_fulfilled

tests/test_order.py
```diff
@@ -1,106 +1,112 @@
...
+            "PLACED",
+            "PLACEMENT_FAILED",
+            "DELIVERY_INITIALIZATION_FAILED",
+            "BEING_FULFILLED",
+            "DOWNLOAD_FAILED",
+            "DOWNLOADED",
+            "FAILED_PERMANENTLY",
+        ],
+    )
+    def test_should_fail_to_get_assets_for_unfulfilled_order(self, status):
```
javidq, line 104: test_fails_to_get_assets_if_not_fulfilled

tests/test_order.py
```diff
@@ -1,106 +1,112 @@
...
+        auth = mock.MagicMock()
+        info = {"status": status}
+        order_obj = order.Order(auth=auth, order_id=constants.ORDER_ID, order_info=info)
+        with pytest.raises(ValueError):
```
javidq, line 108: Let's switch to a modeled ValueError subclass in the code

tests/test_order.py
andher1802: We need to pass auth_mock for calling the get_assets method, and if so it will try to call the info endpoint as well. For clarity I will not pass info in the order initialization.

tests/test_order.py
```diff
@@ -1,106 +1,120 @@
...
+            "PLACED",
+            "PLACEMENT_FAILED",
+            "DELIVERY_INITIALIZATION_FAILED",
+            "BEING_FULFILLED",
+            "DOWNLOAD_FAILED",
+            "DOWNLOADED",
+            "FAILED_PERMANENTLY",
+        ],
+    )
+    def test_fails_to_get_assets_if_not_fulfilled(self, status):
```
javidq, line 112: status type hint

tests/test_order.py
```diff
@@ -1,106 +1,120 @@
...
+        auth = mock.MagicMock()
```
javidq, line 113: We need real auth_mock here, since the status is queried eagerly

tests/test_order.py
```diff
@@ -1,106 +1,120 @@
...
+        info = {"status": status}
+        order_obj = order.Order(auth=auth, order_id=constants.ORDER_ID, order_info=info)
+        with pytest.raises(order.OrderUnfulfilledError):
```
javidq, line 116: `match=f".*{constants.ORDER_ID}.*{status}.*"` is needed to verify dynamic parts

tests/test_order.py
```diff
@@ -1,106 +1,120 @@
...
+                    "id": constants.ASSET_ORDER_ID,
+                }
+            ],
+            "totalPages": 1,
+            "totalElements": 1,
+        }
+        requests_mock.get(url=url_asset_info, json=asset_info)
+        order_obj = order.Order(auth=auth_mock, order_id=constants.ORDER_ID)
+        assets = order_obj.get_assets()
+        assert all(isinstance(asset_element, asset.Asset) for asset_element in assets)
```
javidq, line 95: Why not asset_obj?

javidq: Also can be simplified with `asset_obj, = order_obj.get_assets()`

tests/test_order.py
andher1802: But the fail error is returned before actually call the endpoint.

javidq: No, it calls the mock and fails because the mock still doesn't return the correct status (but not the one we expect it to return)

tests/test_order.py
```diff
@@ -1,106 +1,122 @@
...
+            "FAILED_PERMANENTLY",
+        ],
+    )
+    def test_fails_to_get_assets_if_not_fulfilled(self, auth_mock, requests_mock: req_mock.Mocker, status: str):
+        info = {"status": status}
+        requests_mock.get(url=ORDER_URL, json=info)
+        order_obj = order.Order(auth=auth_mock, order_id=constants.ORDER_ID, order_info=info)
+        with pytest.raises(order.UnfulfilledOrder) as err:
+            order_obj.get_assets()
+        assert re.compile(f".*{constants.ORDER_ID}.*{status}").match(str(err.value)) is not None
```
javidq, line 119: Why not as match in pytest.raises call?

andher1802: you are right, switched to pytest match

andher1802 (approved):

PR #686, opened 2024-11-18
PT-5807 Adding token decoding instead of endpoint retrieval

andher1802:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update


javidq:
Will be addressed in #691

PR #717, opened 2025-01-06
Update Item tags

matheuspinheirosantos:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ You can check your current version with the following command:
 
 For more information, see [UP42 Python package description](https://pypi.org/project/up42-py/).
 ## 2.2.0a20
-**Jan 8, 2025**
+**Jan 9, 2025**
 - Added `update` extension method to `pystac::Item`
 
 ## 2.2.0a19
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ You can check your current version with the following command:
     ```
 
 For more information, see [UP42 Python package description](https://pypi.org/project/up42-py/).
+## 2.2.0a20
+**Jan 8, 2025**
+- Added `update` extension method to `pystac::Item`
+
 ## 2.2.0a19
 **Dec 20, 2024**
 - Switched to using stac client descriptor in `Asset` class and reduced duplication.
--- a/up42/stac.py
+++ b/up42/stac.py
@@ -27,18 +27,19 @@ class FileProvider:
 class UpdateItem:
     session = base.Session()
 
-    def __call__(self, instance):
-        url = host.endpoint(f"/v2/assets/stac/collections/{instance.collection_id}/items/{instance.id}")
-        response = self.session.patch(
-            url=url,
-            json={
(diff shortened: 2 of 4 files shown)


up42/stac.py
```diff
@@ -24,5 +24,25 @@ def __get__(self, obj: Optional[pystac.Asset], obj_type=None) -> Optional[utils.
...
+            ).json()
+        )
+        for key, value in vars(response).items():
+            setattr(item, key, value)
+
+
 def extend():
     pystac.Asset.file = FileProvider()  # type: ignore
+    update_item = UpdateItem()
+    pystac.Item.update = lambda self: update_item(self)  # type: ignore # pylint: disable=unnecessary-lambda
```
javidq, line 48: pystac.Item.update = UpdateItem()

up42/stac.py
```diff
@@ -24,5 +24,25 @@ def __get__(self, obj: Optional[pystac.Asset], obj_type=None) -> Optional[utils.
...
+class UpdateItem:
+    session = base.Session()
+
+    def __call__(self, item):
+        url = host.endpoint(f"/v2/assets/stac/collections/{item.collection_id}/items/{item.id}")
+        response = pystac.Item.from_dict(
+            self.session.patch(
+                url=url,
+                json={
+                    "up42-user:title": item.properties.get("up42-user:title"),
```
javidq, line 36: The key to be extracted to a costant and the same below

up42/stac.py
```diff
@@ -24,5 +24,25 @@ def __get__(self, obj: Optional[pystac.Asset], obj_type=None) -> Optional[utils.
             raise AttributeError
 
 
+class UpdateItem:
+    session = base.Session()
+
+    def __call__(self, item):
+        url = host.endpoint(f"/v2/assets/stac/collections/{item.collection_id}/items/{item.id}")
```
javidq, line 31: is it collection_id or collection?

javidq (changes requested):

up42/stac.py
matheuspinheirosantos: `collection_id` is filled during Item initialization, `collection` can be `str` or `pystac.Collection`

up42/stac.py
matheuspinheirosantos: It doesn't work, I tried. If we do this I don't have a reference to the item in the `__call__` method. I tried doing `update_item()` as a function instead of a class, but I also didn't manage to inject the session in this way. 

This was the only solution I found in the end:
* the `self` in the `UpdateItem` class is still a reference to it, so we can access the session
* we have a reference to the `item` in the `__call__` method.

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ You can check your current version with the following command:
     ```
 
+## 2.2.0a20
+**Jan 8, 2025**
```
javidq, line 33: To be updated

javidq (dismissed):

JeremiahDo (approved):

javidq (approved):

PR #727, opened 2025-03-07
Feat/user 2337 add auth support for sa region

paige-up42:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update


tests/test_host.py
```diff
@@ -19,3 +24,53 @@ def test_should_use_domain_in_token_endpoint_url(domain):
...
+            SA_TOKEN_ENDPOINT,
+        ],
+    ],
+)
+def test_should_use_correct_token_endpoint_url_for_requested_region(region, expected):
+    host.REGION = region
+    assert expected == host.token_endpoint()
+
+
+@pytest.mark.parametrize("region", ["us", "jp"])
```
InesVogel, line 47: NIT: I would use random values here that do not match the expected value instead of values that have a semantic meaning, like done in line 11 and 17

up42/host.py
```diff
@@ -1,4 +1,5 @@
 DOMAIN = "com"
+REGION = "eu"
 
 
 def endpoint(path: str):
```
InesVogel, line 5: does this also need to be changed?

tests/test_host.py
paige-up42: sure thing, done!

up42/host.py
paige-up42: Yes, but since we don't have anything todo with ordering i didn't want to touch this endpoint

up42/base.py
```diff
@@ -46,6 +46,7 @@ def authenticate(
         cfg_file: Optional[Union[str, pathlib.Path]] = None,
         username: Optional[str] = None,
         password: Optional[str] = None,
+        region: Literal["eu", "sa"] = "eu",
```
InesVogel, line 49: if we want our international customers to be able to omit the `region` parameter, we could do 

`region: Optional[Literal["eu", "sa"]] = None,`

and then handle the region = None in the code by setting it to "eu" in case it is None

InesVogel:
generally speaking this looks good, but it would be great to get a review from the former SDK owners (@javidq) as none of us (user team) has worked with the SDK so far.

up42/base.py
jmigueldelgado: with the default set as above the user will be able to omit the `region` parameter already, so we can keep it IMO

up42/host.py
jmigueldelgado: The endpoint refers to any endpoint in the UP42 API, i.e. not only ordering. The user team owns up42/host.py, so you could go ahead and make the change.

jmigueldelgado (commented):
Just one comment wrt endpoint method: the endpoint refers to any endpoint in the UP42 API, i.e. not only ordering. The user team owns up42/host.py, so you could go ahead and make the change.

up42/host.py
paige-up42: adjusted it and now also is region dependant

up42/base.py
paige-up42: i mean it's a question what the default value should be None or eu and i would opt for eu, as this way it's more visible where we point by default

InesVogel: my bad... I thought this would force the user to set the parameter

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ You can check your current version with the following command:
     ```
 
+## 2.2.0a24
+**Mar 13, 2025**
+- Add support for Sa Region.
```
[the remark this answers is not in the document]
jmigueldelgado: We don't have it documented, so I gather [some basic stuff here](https://up42.atlassian.net/wiki/x/B4B2NwE?atlOrigin=eyJpIjoiMTUxMTkzZGFmNDE1NGNlNzg5NmY0OTZhNjdjMjBiY2YiLCJwIjoiYyJ9):
-  start your entries by using the following verbs in the past tense: `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed` 
- if it is a vulnerability fix, start your entry with `Security`
- refer to instance methods, class methods and static methods in the following way: `Class::method` if `Class` is in the `up42` namespace (in `up42/__init__.py`), otherwise `module.Class::method`

jmigueldelgado (approved):
🚀 LGTM!

up42/base.py
```diff
@@ -56,6 +57,7 @@ def authenticate(
             username: The username for the UP42 account (email UP42 console).
             password: Password for the UP42 console login.
         """
```
alessiocol, line 59: a docstring for the new `region` parameter is missing

up42/host.py
```diff
@@ -1,14 +1,31 @@
...
+    if REGION == "eu":
+        return f"https://api.up42.{DOMAIN}{path}"
+    elif REGION == "sa":
+        return f"https://api.sa.up42.{DOMAIN}{path}"
+    raise UnsupportedRegion(f"Region {REGION} is not supported")
 
 
 def token_endpoint():
-    return f"https://auth.up42.{DOMAIN}/realms/public/protocol/openid-connect/token"
+    if REGION == "eu":
```
alessiocol, line 15: what if we use `def endpoint()` inside `token_endpoint()` and `user_info_endpoint()`?
Could simplify both functions to sth like
```
def token_endpoint():
   return endpoint("/realms/public/protocol/openid-connect/token")`
```

paige-up42: i agree, but i would need to introduce the possibility to add an optional domain prefix to `def endpoint()`, which could make it more generic, but prob harder to read.

Afaik, we only have those two endpoints rn that require a different prefix, so i'm open to change it, if that's what we want 🙂

tests/test_host.py
```diff
@@ -2,20 +2,101 @@
 
 from up42 import host
 
+ENDPOINT = "https://api.up42.com/path"
+SA_ENDPOINT = "https://api.sa.up42.com/path"
+TOKEN_ENDPOINT = "https://auth.up42.com/realms/public/protocol/openid-connect/token"
+SA_TOKEN_ENDPOINT = "https://auth.sa.up42.com/realms/public/protocol/openid-connect/token"
+USER_INFO_ENDPOINT = "https://auth.up42.com/realms/public/protocol/openid-connect/userinfo"
+SA_USER_INFO_ENDPOINT = "https://auth.sa.up42.com/realms/public/protocol/openid-connect/userinfo"
```
alessiocol, line 10: I'm not entirely familiar yet with the entire Python SDK, so I might be missing some details :)
Should some of these constants be part of https://github.com/up42/up42-py/blob/master/tests/constants.py ? (I'd have a similar comment also above in `tests/test_base.py`)
For example, I see there is `API_HOST = "https://api.up42.com"` is available and used in other tests.
Perhaps we could simplify this.

up42/host.py
alessiocol: would we need to do that even if `REGION` is a global variable?

up42/base.py
paige-up42: Done 🙂

up42/host.py
alessiocol: forget what I said, I misread 😓

tests/test_host.py
paige-up42: I changed the `Endpoint` to use the const

The rest still gets declared in there, because they are very much specific to this test, because the Token & userinfo use the sub domain `auth.up42`, which is specific to auth and the sa API host only gets used here, so i didn't want to move them to const, premature optimisation and so on

alessiocol:
LGTM

alessiocol (approved):
LGTM

PR #741, opened 2025-05-27
[PT-6368] Deprecate all public methods in the Asset class

jmigueldelgado:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,12 +32,12 @@ For more information, see [UP42 Python package description](https://pypi.org/pro
 
 ## 2.3.0a2
 **May 27, 2025**
-- Deprecated `Asset::get`
-- Deprecated `Asset::all`
-- Deprecated `Asset::stac_items`
-- Deprecated `Asset::stac_info`
-- Deprecated `Asset::file` in favour of `pystac::Asset.file`
-- Deprecated `Asset::download` in favour of `pystac::Asset.file::download`
+- Deprecated `Asset.get`
+- Deprecated `Asset.all`
+- Deprecated `Asset.stac_items` in favour of `pystac::Client.get_items`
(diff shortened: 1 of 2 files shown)


matheuspinheirosantos (approved):

CHANGELOG.md
```diff
@@ -30,6 +30,15 @@ You can check your current version with the following command:
 
 For more information, see [UP42 Python package description](https://pypi.org/project/up42-py/).
 
+## 2.3.0a2
```
dustgalactic, line 33: @jmigueldelgado 
As per semantic versioning, deprecation warrants a new minor version. 
See https://semver.org/
>  issue a new minor release with the deprecation in place

up42/asset.py
```diff
@@ -135,6 +138,7 @@ def stac_items(self) -> pystac.ItemCollection:
             raise ValueError(f"No STAC metadata information available for this asset {self.asset_id}") from exc
 
     @property
+    @utils.deprecation(None, "3.0.0")
```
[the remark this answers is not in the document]
jmigueldelgado: Yes, but instead I would recommend `pystac::Client.get_item`

up42/asset.py
```diff
@@ -114,6 +116,7 @@ def all(
         return map(cls._from_metadata, utils.paged_query(params, "/v2/assets", cls.session))
 
     @property
+    @utils.deprecation("pystac::Client.get_items", "3.0.0")
```
jmigueldelgado, line 119: should we instead recommend `pystac::Client.search`? Because what we suggest in notebooks is search by tags and search by order id. For now users could even use search by asset id.

dustgalactic: Yeah, makes sense.

PR #762, opened 2025-07-30
[PT-5400] add bulk deletion

jmigueldelgado:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/tests/test_stac.py
+++ b/tests/test_stac.py
@@ -172,7 +172,6 @@ class TestBulkDeletion:
     def test_should_raise_and_not_submit_when_missing_items(self):
         mock_stac_client = mock.Mock()
         mock_stac_client.get_items.return_value = iter([self.items[0]])
-
         bulk_deletion = stac.BulkDeletion(self.items[0].id)
         bulk_deletion.stac_client = mock_stac_client
         with pytest.raises(stac.IncompleteCollectionDeletionError):
--- a/tests/test_stac.py
+++ b/tests/test_stac.py
@@ -173,9 +173,8 @@ class TestBulkDeletion:
         mock_stac_client = mock.Mock()
         mock_stac_client.get_items.return_value = iter([self.items[0]])
 
-        bulk_deletion = stac.BulkDeletion()
+        bulk_deletion = stac.BulkDeletion(self.items[0].id)
         bulk_deletion.stac_client = mock_stac_client
-        bulk_deletion.add(self.items[0].id)
         with pytest.raises(stac.IncompleteCollectionDeletionError):
             bulk_deletion.delete()
 
@@ -187,14 +186,8 @@ class TestBulkDeletion:
         mock_stac_client = mock.Mock()
         mock_stac_client.get_items.return_value = iter(self.items)
 
-        bulk_deletion = stac.BulkDeletion()
+        bulk_deletion = stac.BulkDeletion(self.items[0].id, self.items[1].id)
         bulk_deletion.stac_client = mock_stac_client
-        bulk_deletion.add(self.items[0].id, self.items[1].id)
 
         bulk_deletion.delete()
         assert requests_mock.request_history[0].method == "DELETE"
-
-    def test_should_raise_if_no_items_staged(self):
-        bulk_deletion = stac.BulkDeletion()
-        with pytest.raises(ValueError):
-            bulk_deletion.delete()
--- a/tests/test_stac.py
+++ b/tests/test_stac.py
@@ -158,7 +158,7 @@ class TestBulkDeletion:
         )
         for i in range(2)
     ]
-    empty_collection = pystac.Collection(
+    collection_with_all_items = pystac.Collection(
         id="collection-id",
         description="",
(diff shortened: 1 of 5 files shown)


up42/stac.py
```diff
@@ -9,6 +9,12 @@ class InvalidUp42Asset(ValueError):
     pass
 
 
+class IncompleteCollectionDeletionError(ValueError):
```
janchrizz, line 12: are we sure going for this UX?Because this is a bit contradicting from our Console behavior, where when a user selects and tries to delete an item (which will delete other items in the collection), the UI warns the user but still goes ahead with the deletion, instead of blocking it

jmigueldelgado: The SDK is an automation tool, so any confirmation or warning step that we would introduce before deletion would could easily be automated by the user, which could lead to data loss. This was the way we found that would ensure that the user knows exactly what they are deleting.

tests/test_stac.py
```diff
@@ -143,3 +144,74 @@ def test_should_set_up42_extension(self, entity, entity_dict, attribute, key):
...
+            responses = bulk_deletion.submit()
+            assert len(bulk_deletion._collections) == 0  # pylint: disable=protected-access
+            for collection, response in responses.items():
+                assert response.status_code == 204
+                assert collection not in bulk_deletion._collections  # pylint: disable=protected-access
+
+    def test_should_raise_if_no_items_staged(self):
+        bulk_deletion = stac.BulkDeletion()
+        with pytest.raises(ValueError):
+            bulk_deletion.submit()
```
janchrizz, line 217: does it make sense to add 1 additional test for when response.status_code is not 204 from the delete endpoint (for whatever reason)?

up42/stac.py
```diff
@@ -96,3 +102,67 @@ def extend():
...
+        Submit the bulk deletion request. This will effectively delete all items in the collections in the stash.
+
+        Returns:
+            A dictionary mapping collection IDs to their corresponding DELETE responses.
+        """
+        if not self._collections:
+            raise ValueError("No items to delete. Use add() to add item IDs before submitting.")
+
+        if validation_error := self.validate():
+            self._collections.clear()
```
janchrizz, line 157: why are items are not cleared?

up42/stac.py
```diff
@@ -96,3 +102,67 @@ def extend():
...
+        Collections are deleted as a whole: all items in any affected collection must be included.
+
+        Args:
+            *item_ids: Individual item IDs to add for deletion
+        """
+        items = self.stac_client.get_items(*item_ids)
+
+        for item in items:
+            collection = item.get_parent()
+            if isinstance(collection, pystac.Collection):
```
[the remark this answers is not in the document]
jmigueldelgado: Actually no. Fixed

up42/stac.py
```diff
@@ -96,3 +102,67 @@ def extend():
...
+            missing_items = staged_item_ids - set(item.id for item in self._items)
+            if missing_items:
+                error_msg = (
+                    f"Collection '{collection.id}' cannot be deleted because the following items were not included "
+                    f"in the deletion request: {list(missing_items)}. "
+                    f"Collections are deleted as a whole, so all {len(staged_item_ids)} items in collection "
+                    f"'{collection.id}' must be explicitly added for deletion. "
+                    f"Please add the missing items to your deletion request."
+                )
+                return IncompleteCollectionDeletionError(error_msg)
```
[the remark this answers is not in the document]
jmigueldelgado: we don't, fixed

up42/stac.py
jmigueldelgado: since we are now raising the error, we don't need to clear anymore

up42/stac.py
```diff
@@ -96,3 +102,67 @@ def extend():
...
+
+        responses = {}
+        for collection in self._collections:
+            url = host.endpoint(f"/v2/assets/stac/collections/{collection.id}")
+            response = self.session.delete(url=url)
+            responses[collection.id] = response
+
+        self._collections.clear()
+        self._items.clear()
+        return responses
```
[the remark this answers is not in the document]
jmigueldelgado: I was imagining a case where a user wants to delete thousand of items and some might fail. We can also remove it for now and see if users request it. There would be other ways of verifying if a collection was successfully deleted. Dropped it.

dustgalactic:
pyproject.toml update is missing

up42/stac.py
```diff
@@ -96,3 +102,40 @@ def extend():
...
+
+    def add(self, *item_ids: str):
+        items = self.stac_client.get_items(*item_ids)
+
+        for item in items:
+            collection = item.get_parent()
+            self._item_ids.add(item.id)
+            self._collections.add(collection)  # type: ignore
+
+    def delete(self):
```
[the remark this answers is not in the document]
matheuspinheirosantos: Yes, it will, and then it will fail, they should create a new bulk deletion object. I don't see the point of clearing it, that object represents a bulk deletion attempt.

PR #830, opened 2026-06-04
PT-7315 add budgetId for job execution

janchrizz:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/up42/processing_templates.py
+++ b/up42/processing_templates.py
@@ -107,13 +107,13 @@ class JobTemplate:
     def is_valid(self) -> bool:
         return not self.errors
 
-    def execute(self) -> processing.Job:
+    def execute(self, *, budget_id: str | None = None) -> processing.Job:
         url = host.endpoint(
             f"/v2/processing/processes/{self.process_id}/execution"
         )
         job_metadata = self.session.post(
             url,
-            params={"workspaceId": self.workspace_id},
+            params={"workspaceId": self.workspace_id, "budgetId": budget_id},
             json={"inputs": self.inputs},
         ).json()
         return processing.Job.from_metadata(job_metadata)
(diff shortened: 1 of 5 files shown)


up42/processing_templates.py
```diff
@@ -107,13 +107,13 @@ def __evaluate(self):
...
 
-    def execute(self) -> processing.Job:
+    def execute(self, *, budget_id: str | None = None) -> processing.Job:
         url = host.endpoint(
             f"/v2/processing/processes/{self.process_id}/execution"
         )
         job_metadata = self.session.post(
             url,
-            params={"workspaceId": self.workspace_id},
+            params={"workspaceId": self.workspace_id, "budgetId": budget_id},
```
matheuspinheirosantos, line 116: NIT: I would prefer to only pass the param if given

up42/processing_templates.py
```diff
@@ -107,13 +107,13 @@ def __evaluate(self):
     def is_valid(self) -> bool:
         return not self.errors
 
-    def execute(self) -> processing.Job:
+    def execute(self, *, budget_id: str | None = None) -> processing.Job:
```
matheuspinheirosantos: why do we need the `*`?

up42/processing_templates.py
janchrizz: requests will automatically drop params None, so it wont be in the url

up42/processing_templates.py
janchrizz: this is so users can't do execute("id"), and has to do execute(budget_id="id"). but let me remove it

matheuspinheirosantos (approved):

PR #832, opened 2026-06-11
USER-3836: add budget module

JoHeinem:
part of SDK 4.0.0
Querying endpoints:

GET /v2/budgets/{budget-id}

GET /v2/budgets 

GET  /v2/budgets/{budget-id}/usage

ticket : https://up42.atlassian.net/jira/software/projects/USER/boards/84?selectedIssue=USER-3836


up42/budgets.py
```diff
@@ -0,0 +1,87 @@
+import dataclasses
+from collections.abc import Iterator
+from typing import Literal, TypeAlias
+
+from up42 import base, host, utils
+
+logger = utils.get_logger(__name__)
```
anat-o: nit: I think all debug messages here can be removed and this logger is not required

up42/budgets.py
```diff
@@ -0,0 +1,87 @@
...
+
+@dataclasses.dataclass
+class Budget:
+    session = base.Session()
+    id: str
+    name: str
+    status: BudgetStatus
+    created_by: str
+    created_at: str
+    updated_at: str
```
anat-o, line 25: is this field also nullable?

up42/budgets.py
```diff
@@ -0,0 +1,87 @@
...
+            created_by=metadata["createdBy"],
+            created_at=metadata["createdAt"],
+            updated_at=metadata["updatedAt"],
+        )
+
+    @classmethod
+    def get(cls, budget_id: str) -> "Budget":
+        url = host.endpoint(f"/v2/budgets/{budget_id}")
+        logger.debug("Fetching budget %s", budget_id)
+        metadata = cls.session.get(url).json()["data"]
```
anat-o: is ["data"] part of the response object? https://docs.up42.dev/openapi/swagger-ui/?urls.primaryName=payment-service-v2#/Budgets/getBudget 
(same for other methods in this class)

up42/budgets.py
wiemgargouri: nope this fields are not nullable, only `description` and` externalId `are nullable in the  BudgetResponse

up42/budgets.py
wiemgargouri: Great catch! right :) the API doesn't have a `data `wrapper in the responses. fixed !

up42/budgets.py
wiemgargouri: removed

up42/budgets.py
anat-o: ooc, what would be the value of a newly created budget (with no updates yet), would it be equal to created_at?

wiemgargouri: exactly

tests/test_budgets.py
```diff
@@ -0,0 +1,133 @@
...
+                "createdAt": "2026-01-01",
+                "updatedAt": "2026-01-02",
+            },
+            {
+                "id": BUDGET_ID,
+                "name": "some-name",
+                "description": None,
+                "externalId": None,
+                "status": "ACTIVE",
+                "createdBy": "user",
```
janchrizz: nit: createdBy is a uuid, so lets keep the test consistent with the real endpoint

tests/test_budgets.py
```diff
@@ -0,0 +1,133 @@
...
+        assert budgets.Budget.get(BUDGET_ID) == budget
+
+    @pytest.mark.parametrize(
+        "response_metadata",
+        [
+            {
+                "id": BUDGET_ID,
+                "name": "some-name",
+                "status": "ACTIVE",
+                "createdBy": "user",
```
janchrizz: nit: createdBy is a uuid, so lets keep the test consistent with the real endpoint

janchrizz (approved):

anat-o (approved):

tests/test_budgets.py
wiemgargouri: thanks, refactored

dustgalactic (approved):

PR #834, opened 2026-06-16
update order's assigned budgetId

anat-o:
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update

### DIFF
--- a/tests/test_order.py
+++ b/tests/test_order.py
@@ -4,6 +4,7 @@ import uuid
 from typing import Any
 
 import pytest
+import requests
 import requests_mock as req_mock
 
 from tests import constants, helpers
@@ -34,6 +35,7 @@ def _base_order_metadata():
         "type": "ARCHIVE",
         "dataProductId": constants.DATA_PRODUCT_ID,
         "tags": ["some", "tags"],
+        "budgetId": str(uuid.uuid4()),
     }
 
 
@@ -102,6 +104,7 @@ def _base_order(base_order_metadata: dict):
         tags=["some", "tags"],
         info=base_order_metadata,
         details=None,
+        budget_id=str(uuid.uuid4()),
     )
 
 
@@ -447,8 +450,7 @@ class TestOrder:
     def test_update_handles_problem_response(
         self, requests_mock: req_mock.Mocker
     ):
-        budget_id = str(uuid.uuid4())
-        error_response = {
+        mocked_response = {
             "status": 400,
             "title": "Budget cannot be updated because credits have already been deducted",
             "type": "https://docs.up42.com/problems/bad-order-request",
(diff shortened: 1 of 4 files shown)


paras-mathur (approved):

ogheneovieoki-peter-up42 (approved):

tests/test_order.py
```diff
@@ -439,6 +444,69 @@ def test_update_handles_malformed_response(
...
+            "detail": "Can't patch budgetId for an already deducted order",
+        }
+
+        requests_mock.patch(
+            url=ORDER_URL,
+            status_code=400,
+            json=error_response,
+        )
+
+        with pytest.raises(Exception):
```
janchrizz: can we be more specific with the exception? this is too generic an the test may pass (false positive) for unintended errors.
maybe can do something like
```
with pytest.raises(requests.HTTPError) as exc_info:
        order.Order.update(constants.ORDER_ID, budget_id=budget_id)
    assert exc_info.value.response.status_code == 400
    error_message = str(exc_info.value)
    assert stuffs in error_message
```

tests/test_order.py
```diff
@@ -439,6 +444,69 @@ def test_update_handles_malformed_response(
...
+        )
+        expected_order = dataclasses.replace(
+            base_order, budget_id=None, info=updated_metadata
+        )
+        assert (
+            order.Order.update(constants.ORDER_ID, budget_id=None)
+            == expected_order
+        )
+
+    def test_update_without_budget_id_does_not_include_in_body(
```
janchrizz: unnecessary, already covered by `test_should_update`

tests/test_order.py
```diff
@@ -439,6 +444,69 @@ def test_update_handles_malformed_response(
...
+        requests_mock.patch(
+            url=ORDER_URL,
+            status_code=400,
+            json=error_response,
+        )
+
+        with pytest.raises(Exception):
+            order.Order.update(constants.ORDER_ID, budget_id=budget_id)
+
+    def test_update_with_budget_id_set_to_none_removes_budget(
```
janchrizz: not sure what your intention is. but since base_order_metadata doesnt have budgetid in the first place, i think the test is misleading. either seed it with budgetId first, or rename the test

anat-o: I wanted to test that calling order.Order.update with `budget_id = null` will call patch with `budget_id = null` (used to remove the budgetId from the order). i've updated  base_order and base_order_metadata to include the budget_id tho

tests/test_order.py
anat-o: here I wanted to verify, that if budgetId is not sent (as opposed to sent with None), it isn't being sent to the patch api (i.e if i'd send null to patch api, the budgetId would be removed from the order)

tests/test_order.py
anat-o: (additional_matcher=helpers.match_request_body({"budgetId": None})) is basically where I verify what's called to patch

janchrizz (approved):

PR #835, opened 2026-06-18
[USER-4053] deprecate+test: deprecate `workspace_id`

wiemgargouri:
deprecate `workspace_id`
See changelog.

Checklist:
* [x] Test coverage
* [x] Version bump
* [x] Changelog update


up42/processing_templates.py
```diff
@@ -11,7 +11,7 @@
 class JobTemplate:
     session = base.Session()
     process_id: ClassVar[str]
-    workspace_id: str | base.WorkspaceId
+    user_id: str | base.UserId
```
jmigueldelgado: At the moment the user can execute jobs in whatever workspace they wish. With the workspace id removal, we will probably not want to let users run jobs for other users, so the parameter will probably be dropped on the API side. For this PR I propose:
- drop the workspace id from the template
- replace the value of `workspace_id` by the value of `user_id` in the API call
- do not add `user_id` to the JobTemplate

We will align in the team and confirm if this is the way forward.

jmigueldelgado (commented):
See my comment in `up42/processing_templates.py`

up42/processing_templates.py
wiemgargouri: Thanks for the feedback. I refactored the deprecation approach: workspace_id is still accepted and emits a warning, but I did not add user_id. I also added a safeguard for templates that don't define workspace_id.

At the moment, workspace_id remains optional for backward compatibility. We can remove it completely in version 5.0.0 once the deprecation period is over, which avoids breaking existing users during the migration.

paige-up42 (approved):

paige-up42:
LGTM!

up42/order_template.py
```diff
@@ -51,15 +52,21 @@ def _get_items(data: dict, result_type):
...
     data_product_id: str
     display_name: str
     features: geojson.FeatureCollection
     params: dict
+    workspace_id: str | None = None
     tags: list[str] | None = None
     budget_id: str | None = None
 
     def __post_init__(self):
+        if self.workspace_id is not None:
```
javidq, line 64: This message is not needed, since it is done on the descriptor level

up42/order_template.py
```diff
@@ -51,15 +52,21 @@ def _get_items(data: dict, result_type):
 @dataclasses.dataclass
 class BatchOrderTemplate:
     session = base.Session()
-    workspace_id = base.WorkspaceId()
```
javidq, line 54: This is a breaking change, please consider alternatives

javidq (changes requested):

up42/order_template.py
ogheneovieoki-peter-up42, line 54: i agree with Javid here. deprecation should not be removal. we can remove after deprecation

wiemgargouri: could you please elaborate how replacing `workspace_id = base.WorkspaceId() `with `workspace_id: str | None = None` is a breaking change ?

paige-up42: First off. Such comments are not helpful. Please be more mindful on your statement and contribute with constructive feedback. Rather then just criticizing. This helps people better understand what you want from them, and it also helps them learn from it for future contributions.

Secondly, it's not. We moved the default value from the template and into the place method. Thus, the functionality remains the same, just that we can differ if the customer supplied an argument or if they used the fallback.

up42/order_template.py
wiemgargouri: hmm I think you're right if we still use `workspace_id = base.WorkspaceId()`. 

The warning in __post_init__ was added because I had replaced:

`workspace_id = base.WorkspaceId()` with `workspace_id: str | None = None`

After that change, workspace_id was no longer using the WorkspaceId descriptor, so the descriptor-level warning was no longer triggered. I added the warning in __post_init__ to preserve the deprecation notice when users explicitly passed workspace_id.

But if we keep the descriptor-based implementation during the deprecation period, then the warning in __post_init__ is indeed redundant and can be removed. 

does it make sense?

up42/base.py
```diff
@@ -85,12 +85,27 @@ def __get__(self, obj, obj_type=None) -> requests.Session:
 
 
 class WorkspaceId:
```
jmigueldelgado, line 87: You could use the`@utils.deprecation` decorator instead.

jmigueldelgado (commented):
With this usage of `dataclasses.field` 
```suggestion
        workspace_id: str | base.WorkspaceId = dataclasses.field(
        default=base.WorkspaceId()
    )
```
 throughout the code, you will be able to drop all the warnings by simply adding the `@utils.deprecation` to the `base.WorkspaceId` class

wiemgargouri:
closed this PR for now and We will bring it back as part of SDK 4.1.0.

